PR #15, opened 2017-05-25
ComposedFunc

VsSekorin:
Create func package and ComposedFunc.


0crat:

yegor256:
@VsSekorin why only two encapsulated functions? Why not a collection of them? Would be much more convenient.

VsSekorin:
@yegor256 should be
```
private final Iterable<Func<Object, Object>> funcs;
private final Func<Object, Y> last;
```
because I have to return Y. I'm not sure that's a good thing.

PR #20, opened 2017-05-25
Combined LengthOf

englishman:
combine LengthOfIterable and LengthOfIterator into one LengthOf with ctors.


0crat:
@englishman I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

src/main/java/org/cactoos/list/LengthOf.java
```diff
@@ -44,9 +45,17 @@
 
     /**
      * Ctor.
+     * @param items The array
+     */
+    public LengthOf(final Iterable<?> items) {
+        this(items.iterator());
```
yegor256: @englishman this is against one of our fundamental principles: no code in constructors

yegor256:
@englishman see my comment above

src/main/java/org/cactoos/list/LengthOf.java
englishman: @yegor256 pls, be more specific: it's possible to have many ctors and use one of them as a primary ctor. So secondary ctors call the primary with some code: `Elegant Objects` book val. 1 `1.2 Make one ctor primary`, page 29. Here is Cash example with just some code in secondary ctor.

yegor256: @englishman yes, true, there is an example in the book. And it's wrong. I will go through the book and try to find similar examples and will fix them. Thanks for catching that. We must not do anything inside ctors, except assigning private attributes. It's important.

yegor256: @englishman that's why, I don't think it's possible to have just one class `LengthOf`. We will need `LengthOfIterable`, `LenghtOfInput`, `LengthOfIterator`, etc.

englishman: @yegor256 the whole conception of secondary ctors is wrong then as them are useless.

Anyway, here we have 2 options: either apply PR as is (and you don't want it) or cancel it at all as it's impossible to rename both `LengthOfIterable` and `LengthOfIterator` to class with the same name `LengthOf`. WDYT?

englishman: @yegor256 BTW how about these lines
https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/text/Sprintf.java#L59
and
https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/text/Sprintf.java#L69
Here is code in ctors :smile: 

I guess you have to clearly define what kind of code is acceptable in ctors in addition of simple assignments.

yegor256: @englishman I think we have to kill this PR. And you're absolutely right, that code is wrong, but can easily be fixed with `new`. Can you submit PRs for them?

yegor256:
@englishman I don't think it's doable. WDYT?

englishman:
@yegor256 agreed! Closing this PR

PR #39, opened 2017-05-26
IterableAsList

g4s8:
`IterableAsList` to make a list from iterable.


0crat:

src/main/java/org/cactoos/list/IterableAsList.java
```diff
@@ -0,0 +1,103 @@
...
+    /**
+     * Ctor.
+     *
+     * @param iterable An {@link Iterable}
+     */
+    public IterableAsList(final Iterable<T> iterable) {
+        super();
+        this.origin = Collections.list(
+            new IteratorAsEnumeration<>(
+                iterable.iterator()
```
yegor256: @g4s8 this is not acceptable. We can't have any code inside ctors. You're calling `.iterator()`

yegor256:
@g4s8 good idea, but see my comment above

yegor256:
@g4s8 also, I would put that `Repeat` into a different PR

src/test/java/org/cactoos/list/IterableAsListTest.java
```diff
@@ -0,0 +1,110 @@
...
+     *
+     * @throws Exception if failed
+     */
+    @Test
+    public void elementAtIndexTest() throws Exception {
+        // @checkstyle MagicNumber (7 lines)
+        MatcherAssert.assertThat(
+            new IterableAsList<>(
+                new ArrayAsIterable<>(
+                    1, 2, 3, 4
```
yegor256: @g4s8 it's not obvious that number `4` in the array must be exactly the same as the number `4` two lines below. It's better to introduce a variable for them.

src/test/java/org/cactoos/list/IterableAsListTest.java
```diff
@@ -0,0 +1,110 @@
...
+     * Test size of list.
+     *
+     * @throws Exception if failed
+     */
+    @Test
+    public void sizeTest() throws Exception {
+        // @checkstyle MagicNumber (5 lines)
+        MatcherAssert.assertThat(
+            new IterableAsList<>(
+                Collections.nCopies(42, 0)
```
yegor256: @g4s8 same here

yegor256:
@g4s8 a few comments above

src/main/java/org/cactoos/list/IterableAsList.java
```diff
@@ -0,0 +1,107 @@
...
+        this.source = iterable;
+        this.cache = new ArrayList<>(0);
+        this.length = new LengthOfIterable(iterable);
+    }
+
+    @Override
+    public T get(final int index) {
+        if (index < 0 || index >= this.size()) {
+            throw new IndexOutOfBoundsException(
+                String.format(
```
DronMDF: Maybe it's better to use cactoos.text.FormattedText?

g4s8:
@yegor256, @DronMDF All done!

src/test/java/org/cactoos/list/IterableAsListTest.java
```diff
@@ -0,0 +1,112 @@
...
+     *
+     * @throws Exception if failed
+     */
+    @Test
+    public void elementAtIndexTest() throws Exception {
+        final int expected = 4;
+        final int position = 3;
+        final Integer[] array = Collections.nCopies(position + 1, 0)
+            .toArray(new Integer[0]);
+        array[position] = expected;
```
yegor256: @g4s8 we are trying to make all tests single statement. There should be nothing but variables declaration and `assertThat`. I would refactor it like this:

```java
int num = 345;
int[] array = {0, 1, 2, num, 3, 4};
assertThat(IterableAsList(...).get(3), equalTo(num))
```

yegor256:
@g4s8 see above

g4s8:
@yegor256 fixed

yegor256:
@rultor merge

PR #40, opened 2017-05-27
Repeat iterable

g4s8:
`Repeat` iterable to repeat some element multiple times.

### DIFF
--- a/src/main/java/org/cactoos/list/Repeat.java
+++ b/src/main/java/org/cactoos/list/Repeat.java
@@ -1,0 +1,85 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


0crat:

src/test/java/org/cactoos/list/RepeatTest.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * Test empty 'repeat' size.
+     *
+     * @throws Exception if failed
+     */
+    @Test
+    public void emptyTest() throws Exception {
+        // @checkstyle MagicNumber (5 lines)
```
yegor256: @g4s8 zero is not a magic number, there is no need to suppress Checkstyle here

src/test/java/org/cactoos/list/RepeatTest.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    @Test
+    public void allSameTest() throws Exception {
+        // @checkstyle MagicNumber (11 lines)
+        MatcherAssert.assertThat(
+            new LengthOfIterable(
+                new FilteredIterable<>(
+                    new Repeat<>(
+                        11,
+                        42
```
yegor256: @g4s8 I would move this to a new variable. Now you have 42 two times in this method and it's not really obvious that they are related. This duplication creates coupling between them, which is difficult to trace in the future.

yegor256:
@g4s8 see my comments above

yegor256:
@rultor try to merge

yegor256:
@g4s8 thanks!

src/main/java/org/cactoos/list/Repeat.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @Override
+    public Iterator<T> iterator() {
+        return new RepeatIterator();
+    }
+
+    /**
+     * An iterator.
+     */
+    private final class RepeatIterator implements Iterator<T> {
```
fabriciofx: @g4s8 @yegor256 IMHO, I think private methods *and* classes a mistake. The argument is similar to [it](http://www.yegor256.com/2017/02/07/private-method-is-new-class.html). WDYT?

g4s8: @fabriciofx I think you're right in general, but in this case (iterable+iterator) it's ok to use private classes, because we are forced by jkd architecture to implement iterators for each iterable, when iterator on its own not very useful class, it can be used only with iterable together.

src/main/java/org/cactoos/list/Repeat.java
```diff
@@ -0,0 +1,85 @@
...
+
+        /**
+         * Current position.
+         */
+        private int cursor;
```
g4s8: @fabriciofx I agree, if you know better solution let's submit a PR, but keep in mind that iteration can be performance critical section for library user.

PR #42, opened 2017-05-27
Format string with locale

g4s8:
Introduced `StringFormat` for `FormattedText` with different implementations:
 - `LocalizedFormat` to format text with specific `Locale`
 - `SimpleFormat` that uses `Locale.getDefault(Locale.Category.FORMAT)` to format string


0crat:

src/main/java/org/cactoos/text/StringFormat.java
```diff
@@ -0,0 +1,70 @@
...
+     *
+     * @param arguments Arguments
+     * @return Formatted string
+     */
+    String apply(final Collection<?> arguments);
+
+    /**
+     * Default decorator.
+     */
+    abstract class AbstractFormat implements StringFormat {
```
yegor256: @g4s8 let's stay away from implementation inheritance: c0d03d4

yegor256:
@g4s8 why can't we just add a new ctor to `FormattedText` with `Locale`? Looks like an over-design to me...

g4s8:
@yegor256 
> why can't we just add a new ctor to FormattedText

I think we may need this `LocalizedFormat` later, e.g. to implement `NumberAsString`, because `float` or `double` has different separators (`.`, `,`) in some locales.

yegor256:
@g4s8 well, it's better to wait until we really need it. At the moment it looks like over-design.

g4s8:
@yegor256 reverted `StringFormat`. Btw do you think abstract classes with all final methods are evel too?  I've seen many of them in your projects, like this one https://github.com/yegor256/takes/blob/19a530fe7fb20d801171a1db2e8bdc8d77b24591/src/main/java/org/takes/facets/fork/FkWrap.java

yegor256:
@rultor merge

yegor256:
@g4s8 yes, I think they are evil too. Maybe this is too extreme, but in this library let's try to stay away from them, as much as we can.

yegor256:
@g4s8 thanks!

Issue #44, opened 2017-05-28
Function result


VsSekorin:
Let's introduce new class Result (lazy evaluation). This class is necessary for https://github.com/yegor256/cactoos/issues/43.

0crat:
@vssekorin I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

PR #45, opened 2017-05-28
#44 Result

VsSekorin:
As per #44 
Result introduced

### DIFF
--- a/src/main/java/org/cactoos/func/Result.java
+++ b/src/main/java/org/cactoos/func/Result.java
@@ -21,25 +21,56 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
  * SOFTWARE.
  */
-package org.cactoos;
+package org.cactoos.func;
+
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


0crat:

src/main/java/org/cactoos/func/Result.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    private final Func<X, Y> func;
+    /**
+     * The argument.
+     */
+    private final X arg;
+    /**
+     * The result.
+     */
+    private final List<Y> value;
```
dusan-rychnovsky: Why a list if you only ever need the first element?

VsSekorin: This is caching. Most likely it must be carried out in the decorator. But it seems to me that the object does not make sense without it.
How to implement caching differently with immutable objects?

yegor256:
@VsSekorin I don't think I like the design here. We already have `StickyFunc`. We have to reuse it. How about this:

```java
new FuncAsValue(
  new StickyFunc(func),
  arg
);
```

This `FuncAsValue` will `implement Func.Value`.

Issue #63, opened 2017-05-30
Let's do a not procedural temporary file creation/deletion


fabriciofx:
This [kind of code](https://github.com/yegor256/cactoos/blob/c4f228d1eeef7e165a670b5a5536b398dc2fb5a5/src/test/java/org/cactoos/io/FileAsInputTest.java#L51-L52) is very procedural:
```java
final File temp = File.createTempFile("cactoos-1", "txt-1");
temp.deleteOnExit();
```
So, let's do a not procedural temporary file creation/deletion.

0crat:
@fabriciofx I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

englishman:
We can use [TemporaryFolder](http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html) JUnit rule for this. Specifically, [newFile](http://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html#newFile(java.lang.String)) method.

fabriciofx:
@englishman Do we **must** declare `TemporaryFolder` as a `@Rule`? If yes, I don't like it a good solution, because [tests methods must share nothing](http://www.yegor256.com/2016/05/03/test-methods-must-share-nothing.html).

englishman:
@fabriciofx yes, I know (actually, I'm fighting against this sharing in my current project). Rules are the standard JUnit way to do smth BEFORE and AFTER a test method (grant/free resource etc.).

Otherwise, we have to impl smth like this
```
@Test void testMethod() {
    try(File tmp = new TempFile(...)) {
      ...do testing...
    }
}
```
Might be it's good idea to implement our own `TempFile` class impl.

fabriciofx:
@yegor256 WDYT about this issue? Should I close it or continue? If I should continue, could you assign it to me?

yegor256:
@fabriciofx let's keep it open, we'll find someone to solve it

0crat:
Job [#63](https://github.com/yegor256/cactoos/issues/63) is now in scope, role is `DEV`

0crat:
[...]

PR #65, opened 2017-05-31
#63 Let's do a not procedural temporary file creation/deletion

fabriciofx:


0crat:

src/main/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            if (!Files.deleteIfExists(this.toPath())) {
+                throw new IOException(
+                    new FormattedText(
+                        "file '%s' not deleted because it is not exist",
+                        this.getPath()
+                    ).asString()
+                );
+            }
+        } catch (final DirectoryNotEmptyException ex) {
```
englishman: Can we use `catch` with multuple exceptions?
http://www.journaldev.com/629/java-catch-multiple-exceptions-rethrow-exception

src/main/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,72 @@
...
+ * @version $Id$
+ * @since 0.1
+ */
+@SuppressWarnings("serial")
+public final class TempFile extends File implements AutoCloseable {
+    /**
+     * Ctor.
+     * @param pathname The pathname of the file
+     */
+    public TempFile(final String pathname) {
```
englishman: Do we really need an exact file name? Basically, we don't care about exact temp file name. Let's introduce ctor with directory and _may be_ extension here. Later a client can simply get exact fine name with `File#get...` method call.

src/main/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,72 @@
...
+ * Create an autocloseable temporary file that deletes itself.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.1
+ */
+@SuppressWarnings("serial")
+public final class TempFile extends File implements AutoCloseable {
```
englishman: I'm thinking... _is_ `TempFile` actually `File` successor?

It looks like we need some functionality for temp. file name generation and code for deleting the file at the end of work. Isn't a better design to make `TempFile` _contains_ `File` instance inside? Also additional point to this idea: looks like we don't have a place for temp. file name generation inside of current `TempFile` ctor.

With the new design `TempFile` API may looks like
```
class TempFile implements AutoClosable {

    TempFile(String directory, String mask) { ... }

    File file() {
         ... generate temp file name....
         ... create and store File instance...
        return File instance
    }

   public close() {
      ... remove File instance...
   }
}
```

WDYT?

fabriciofx: @englishman I agree totally with you. Your suggestion is better. I'll review my code soon.

src/main/java/org/cactoos/io/TempFile.java
fabriciofx: @englishman Yeah, again, your suggestion is perfect. Thanks!

src/main/java/org/cactoos/io/TempFile.java
fabriciofx: @englishman I know it, but I thought this code was better because is more explicit. What do you think? If you think better put all exceptions in one line I can do it, no problems! Again, thanks for your review and suggestions!

src/main/java/org/cactoos/io/TempFile.java
englishman: @fabriciofx thank you! Unfortunately I'm still not sure what is a better name than `TempFile` for the class keeping in mind we also can have here e.g. temp. directory creation functionality. May be `TempLocation` or `TempPlace`?

src/main/java/org/cactoos/io/TempFile.java
englishman: @fabriciofx well... I always prefer more expressive and laconic solutions. So I would use here new Java 7 syntax.

fabriciofx: @englishman Ok, I change the code to use Java 7 syntax. Thanks!

yegor256:
@fabriciofx I think we need to have the temporary file creating functionality here as well. At the moment it's just a file closer. We can leave it as is, but need to rename it to something like `DeletedOnCloseFile`.

src/main/java/org/cactoos/io/TempFile.java
vivekimsit: @englishman it doesn't feels right that generating file name is the responsibility of the file instance. There should be some random number generator object for that, wdyt?

englishman: @vivekimsit I think you are right. Later I found out the same idea. Moreover, my `TempFile` proposition with `file` method is mutable...

So, @fabriciofx I think we have to create two entities:
1. temp. file/folder name `implements Text`
2. class as @yegor256 advised [here](https://github.com/yegor256/cactoos/pull/65#issuecomment-305472821)

fabriciofx:
@yegor256 Ok. So I'll close this PR and open other to rename it to something like `DeletedOnCloseFile`. WDYT?

yegor256:
@fabriciofx sure

PR #71, opened 2017-06-01
String as number

g4s8:
`new StringAsNumber("42")` instead of `Integer.parseInt("42")`


0crat:

src/test/java/org/cactoos/text/StringAsNumberTest.java
```diff
@@ -0,0 +1,86 @@
...
+ * @author Kirill (g4s8.public@gmail.com)
+ * @version $Id$
+ * @since 0.2
+ */
+public final class StringAsNumberTest {
+
+    /**
+     * Test {@link Integer}.
+     */
+    @Test
```
yegor256: @g4s8 see how I've done other tests: https://github.com/yegor256/cactoos/blob/master/src/test/java/org/cactoos/io/InputAsBytesTest.java I removed JavaDoc and added "reasons" to `assertThat`. I believe this is a better design. Also, see: https://github.com/teamed/qulice/issues/848

src/main/java/org/cactoos/text/StringAsNumber.java
```diff
@@ -0,0 +1,71 @@
...
+package org.cactoos.text;
+
+/**
+ * String as {@link Number}.
+ *
+ * @author Kirill (g4s8.public@gmail.com)
+ * @version $Id$
+ * @since 0.2
+ */
+public final class StringAsNumber extends Number {
```
yegor256: @g4s8 maybe `TextAsInteger` would be a more logical design in our case?

```java
int x = new TextAsInteger(new StringAsText("-158")).asValue()
```

yegor256:
@g4s8 see some comments above

src/main/java/org/cactoos/text/StringAsNumber.java
yegor256: @g4s8 Also, we'll need `TextAsDouble`, `TextAsBoolean`, `TextAsLong`. Or maybe let's rename this one to `TextAsLong` and let users cast from `long` to `int` after parsing?

g4s8: @yegor256 I think that casting is not elegant solution, so I've added `TextAsLong` and `TextAsInt` as different classes. Furthermore `Long` and `Integer` have separated reference cache for small values (`Long.LongCache.cache` and `Integer.IntegerCache.cache`)

g4s8:
@yegor256 added `Scalar<>` implementations for `Long`, `Integer`, `Double`, `Float` and `Boolean`, fixed tests

yegor256:
@rultor try to merge

PR #73, opened 2017-06-02
Replaced Text

memoyil:
Replaces the text

### DIFF
--- a/src/main/java/org/cactoos/text/ReplacedText.java
+++ b/src/main/java/org/cactoos/text/ReplacedText.java
@@ -43,29 +43,29 @@ public final class ReplacedText implements Text {
     /**
      * The old char.
      */
-    private final String oldchar;
+    private final String needle;
 
     /**
      * The new char.
      */
-    private final String newchar;
+    private final String replacement;
 
     /**
      * Ctor.
      * @param text The text
-     * @param oldchar The old char
-     * @param newchar The new char
+     * @param find The find one
+     * @param replace The replace one
      */
-    public ReplacedText(final Text text, final String oldchar, final String
-        newchar) {
+    public ReplacedText(final Text text, final String find, final String
+        replace) {
         this.origin = text;
-        this.oldchar = oldchar;
-        this.newchar = newchar;
+        this.needle = find;
+        this.replacement = replace;
     }
 
     @Override
     public String asString() throws IOException {
-        return this.origin.asString().replace(this.oldchar, this.newchar);
+        return this.origin.asString().replace(this.needle, this.replacement);
     }
 }
 
--- a/src/main/java/org/cactoos/text/ReplacedText.java
+++ b/src/main/java/org/cactoos/text/ReplacedText.java
@@ -31,7 +31,7 @@ import org.cactoos.Text;
  *
  * @author Mehmet Yildirim (memoyil@gmail.com)
  * @version $Id$
- * @since 0.1
+ * @since 0.2
  */
 public final class ReplacedText implements Text {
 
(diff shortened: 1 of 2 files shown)


0crat:
@memoyil I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

src/main/java/org/cactoos/text/ReplacedText.java
```diff
@@ -0,0 +1,48 @@
...
+    private String newChar;
+
+    /**
+     * Ctor.
+     *
+     * @param text    The text
+     * @param oldChar The old char
+     * @param newChar The new char
+     */
+    public ReplacedText(final Text text, String oldChar, String newChar) {
```
yegor256: @memoyil I don't think the names are good here. I would use `needle` and `replacement` as attribute names. And `find`, `replace` as argument names.

yegor256:
@memoyil see my comment above.

src/main/java/org/cactoos/text/ReplacedText.java
```diff
@@ -0,0 +1,71 @@
...
+
+import java.io.IOException;
+import org.cactoos.Text;
+
+/**
+ * Replace the Text.
+ *
+ * @author Mehmet Yildirim (memoyil@gmail.com)
+ * @version $Id$
+ * @since 0.1
```
yegor256: @memoyil 0.1 is already released. Put 0.2 here.

yegor256:
@memoyil one more small fix and we're ready to go

src/main/java/org/cactoos/text/ReplacedText.java
dusan-rychnovsky: I wonder if there is any way to have this check automated? Checking the version manually every time seems tedious and error prone.

src/test/java/org/cactoos/text/ReplacedTextTest.java
```diff
@@ -0,0 +1,24 @@
...
+ * Test case for {@link ReplacedText}.
+ * @author Mehmet Yildirim (memoyil@gmail.com)
+ * @version $Id$
+ * @since 0.1
+ * @checkstyle JavadocMethodCheck (500 lines)
+ */
+public final class ReplacedTextTest {
+
+    @Test
+    public void replaceText() {
```
memoyil: Added extra test cases for that.

memoyil: I think replaced all text with given value should be another class.

dusan-rychnovsky: Then we should have here a test case that verifies that if you have multiple occurences of the substring in your original string, only the first one gets replaced.

memoyil: You are incorrect because both replace() and replaceAll() replaces all the occurrence found. However, there is difference in the type of arguments they accept.

replace() works on CharSequence, where as replaceAll(), accepts regex, so you can do a regular expression based pattern matching.

memoyil: What i mean is we can add for example ReplacedRegexText or something like that later.

dusan-rychnovsky: I don't understand - I am saying we need a test case here that verifies that this class works as expected when there are multiple occurences of the needle in the original string.

memoyil: Ok i will add the test case. You said that "if you have multiple occurences of the substring in your original string only the first one gets replaced" . So I replied your comment based on that. The replace() function we used, replace all the matchings also.

yegor256:
@rultor try to merge

Issue #49, opened 2017-05-29
StringUtils methods into org.cactoos.text


yegor256:
Let's create classes for all methods from Apache `StringUtils`.

fabriciofx:
@yegor256 Are you doing it or anyone could try it?

englishman:
@fabriciofx I think we should start from adding some issues with specific class names for new functionality to split work then between developers.

fabriciofx:
@englishman I agree. Could you start? :)

yegor256:
@fabriciofx @englishman go ahead guys, I'm not touching this.

englishman:
@fabriciofx i'm on a small vacation. Will come back on this on Monday.

fabriciofx:
@englishman Ok. Do you mind if I advance with some code? And have a nice vacation! :)

englishman:
@fabriciofx absolutely no :+1: go ahead!

fabriciofx:
@englishman 👍

fabriciofx:
@yegor256 I think you can close this issue, because these commits have been opened as individual issues. WDYT?

yegor256:
@fabriciofx right, closing

PR #80, opened 2017-06-03
#49: StringUtils methods into org.cactoos.text

fabriciofx:
As per #49:

- `AbbreviatedText` introduced
- `SplitText` introduced
- `JoinedText` introduced
- `CapitalizedText` introduced
- `TruncatedText` introduced
- `SubText` introduced
- `RepeatedText` introduced
- `NormalizedText` introduced

### DIFF
--- a/src/main/java/org/cactoos/text/AbbreviatedText.java
+++ b/src/main/java/org/cactoos/text/AbbreviatedText.java
@@ -33,7 +33,7 @@ import org.cactoos.Text;
  *
  * @author Fabricio Cabral (fabriciofx@gmail.com)
  * @version $Id$
- * @since 0.2
+ * @since 0.3
  */
 public final class AbbreviatedText implements Text {
 
--- a/src/main/java/org/cactoos/text/AbbreviatedText.java
+++ b/src/main/java/org/cactoos/text/AbbreviatedText.java
@@ -1,0 +1,93 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.text;
+
+import java.io.IOException;
+import org.cactoos.Text;
+
+/**
+ * Abbreviates a Text using ellipses.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.2
+ */
+public final class AbbreviatedText implements Text {
+
+    /**
+     * The text.
+     */
+    private final Text origin;
+
+    /**
+     * The initial position in the text.
+     */
+    private final int offset;
+
+    /**
+     * The max width of the abbreviated text.
+     */
+    private final int width;
+
+    /**
+     * Ctor.
+     * @param text The Text
+     */
+    public AbbreviatedText(final Text text) {
+        // @checkstyle MagicNumber (1 line)
+        this(text, 80);
+    }
+
+    /**
+     * Ctor.
+     * @param text The Text
+     * @param width Width of the result string
+     */
+    public AbbreviatedText(final Text text, final int width) {
+        // @checkstyle MagicNumber (1 line)
+        this(text, 0, width);
+    }
+
+    /**
+     * Ctor.
+     * @param text The Text
+     * @param offset Text position where to start
+     * @param width Width of the result string
+     */
+    public AbbreviatedText(final Text text, final int offset, final int width) {
+        this.origin = text;
+        this.offset = offset;
+        this.width = width;
+    }
+
+    @Override
+    public String asString() throws IOException {
+        return new FormattedText(
+            "%s...",
+            this.origin.asString().substring(this.offset, this.width)
+        ).asString();
+    }
+}
--- a/src/main/java/org/cactoos/text/CapitalizeWordFunc.java
+++ b/src/main/java/org/cactoos/text/CapitalizeWordFunc.java
@@ -23,40 +23,26 @@
  */
 package org.cactoos.text;
 
-import java.io.IOException;
+import org.cactoos.Func;
 import org.cactoos.Text;
-import org.cactoos.list.TransformedIterable;
 
 /**
- * Capitalize the Text.
+ * Func that capitalize a word.
+ *
+ * <p>There is no thread-safety guarantee.
  *
  * @author Fabricio Cabral (fabriciofx@gmail.com)
  * @version $Id$
  * @since 0.3
  */
-public final class CapitalizedText implements Text {
-
-    /**
-     * The text.
-     */
-    private final Text origin;
-
-    /**
-     * Ctor.
-     * @param text A Text
-     */
-    public CapitalizedText(final Text text) {
-        this.origin = text;
-    }
-
+public final class CapitalizeWordFunc implements Func<Text, Text> {
     @Override
-    public String asString() throws IOException {
-        return new JoinedText(
-            new TransformedIterable<Text, Text>(
-                new SplitText(this.origin),
-                new CapitalizeWordFunc()
-            )
-        ).asString();
+    public Text apply(final Text input) throws Exception {
+        return new FormattedText(
+            "%s%s",
+            input.asString().substring(0, 1).toUpperCase(),
+            input.asString().substring(1)
+        );
     }
-}
 
+}
--- a/src/main/java/org/cactoos/text/CapitalizedText.java
+++ b/src/main/java/org/cactoos/text/CapitalizedText.java
@@ -53,15 +53,15 @@ public final class CapitalizedText implements Text {
     @Override
     public String asString() throws IOException {
         return new JoinedText(
-            new TransformedIterable<String, Text>(
+            new TransformedIterable<Text, Text>(
                 new SplitText(this.origin),
-                new Func<String, Text>() {
+                new Func<Text, Text>() {
                     @Override
-                    public Text apply(final String input) throws Exception {
+                    public Text apply(final Text input) throws Exception {
                         return new FormattedText(
                             "%s%s",
-                            input.substring(0, 1).toUpperCase(),
-                            input.substring(1)
+                            input.asString().substring(0, 1).toUpperCase(),
+                            input.asString().substring(1)
                         );
                     }
                 }
--- a/src/main/java/org/cactoos/text/CapitalizedText.java
+++ b/src/main/java/org/cactoos/text/CapitalizedText.java
@@ -1,0 +1,72 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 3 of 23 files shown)


0crat:

src/main/java/org/cactoos/text/SplitedText.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Split a Text.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.3
+ */
+public final class SplitedText implements Iterable<String> {
```
fabriciofx: @yegor256 In your opinion, should `SplitedText` implements an `Iterable<String>` or a `Iterable<Text>`?

dusan-rychnovsky: Should be `SplitText` - split is irregular.

fabriciofx: @dusan-rychnovsky I'll correct it. Thanks!

src/main/java/org/cactoos/text/AbbreviatedText.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    private final int width;
+
+    /**
+     * Ctor.
+     * @param text The Text
+     */
+    public AbbreviatedText(final Text text) {
+        // @checkstyle MagicNumber (1 line)
+        this(text, 80);
```
Timmeey: Why 80? 
Please not use MagicNumber.

src/main/java/org/cactoos/text/SplitText.java
```diff
@@ -0,0 +1,102 @@
...
+    public SplitText(final Text text, final String pattern, final int limit) {
+        this.origin = text;
+        this.pattern = pattern;
+        this.limit = limit;
+    }
+
+    @Override
+    public Iterator<String> iterator() {
+        try {
+            return Arrays.stream(
```
Timmeey: Why not just do
` return new ArrayAsIterable(
  this.origin.asString.split(......)
  ).iterator()`

src/main/java/org/cactoos/text/SplitedText.java
Timmeey: I would strongly argue that we stay as long in the framework type-system as possible, therefor i think it should be Iterable<Text>

src/main/java/org/cactoos/text/AbbreviatedText.java
Timmeey: Okay nevermind. It complains if you define a constant and only use it once....

fabriciofx: @Timmeey Because the old terminals supports at max 80 chars per column. But feel free to suggest another width. 😄

src/main/java/org/cactoos/text/SplitText.java
fabriciofx: @Timmeey Much better! Thanks!

src/main/java/org/cactoos/text/SplitedText.java
fabriciofx: @Timmeey I agree, but an `Iterable` of `String` or of `Text`?

Timmeey: Haha, it cut off due to my usage of < chars.

I meant 
`Iterable<Text>`

src/test/java/org/cactoos/text/SplitTextTest.java
```diff
@@ -0,0 +1,62 @@
...
+public final class SplitTextTest {
+
+    @Test
+    public void splitsText() throws IOException {
+        MatcherAssert.assertThat(
+            "Can't split a text",
+            new SplitText(
+                new StringAsText("one two three four five")
+            ),
+            Matchers.contains("one", "two", "three", "four", "five")
```
fabriciofx: @Timmeey To change `public final SplitText implements Iterable<String>` to `public final SplitText implements Iterable<Text>` we need change the `Matchers.contains("one", "two", "three", "four", "five")` to             `Matchers.contains(new StringAsText("one"), new StringAsText("two"), new StringAsText("three"), new StringAsText("four"),  new StringAsText("five"))`. The problem is to this works, I need implements a method `equals()` into `Text` or `StringAsText`, and a `equals()` method requires an `instanceof` operator, and this projects is against it usage. So, what do you recommend?

Timmeey:
The instanceof in an equal method is a requirement by the language. Therefore arguments to avoid using it don't apply in my opinion. 

The important thing here is to identify what defines the identity of the object, instead of comparing its state. Take a look here http://www.yegor256.com/2014/12/09/immutable-object-state-and-behavior.html


About using an interable of text instead string. It just seems weird to my why the join on texts should produce a text, and the split would produce strings. So I feel. Pretty strongly that producing an interable of text is the right way to go, even though it requires some additions to other parts 

Am 6. Juni 2017 13:58:05 MESZ schrieb Fabricio Cabral <notifications@github.com>:

fabriciofx:
@Timmeey You're totally right. Could you check the new modifications? Thanks in advance!

src/main/java/org/cactoos/text/CapitalizedText.java
```diff
@@ -53,15 +53,15 @@ public CapitalizedText(final Text text) {
...
     public String asString() throws IOException {
         return new JoinedText(
-            new TransformedIterable<String, Text>(
+            new TransformedIterable<Text, Text>(
                 new SplitText(this.origin),
-                new Func<String, Text>() {
+                new Func<Text, Text>() {
                     @Override
-                    public Text apply(final String input) throws Exception {
+                    public Text apply(final Text input) throws Exception {
```
Timmeey: Maybe i'm a bit picky here, but it took me about 30 seconds to really understand what this func does. Maybe we can externalize this into their own private method named capitalizeWord() or use a variable for the function, that is named something like that.

fabriciofx: @Timmeey I agree totally with you, but I'm totally against private classes or methods. What do you think I create a new class called `CapitalizeTextFunc` and instantiate an object instead of this func?

fabriciofx:
@Timmeey Could you check the new modifications? I think the code is get simpler with them!

src/main/java/org/cactoos/text/CapitalizedText.java
Timmeey: Actually i'm against making this a new class, since there is no reuse situation for that class then. 

But why no private methods?

fabriciofx: @Timmeey Well, I think private classes or methods hide functionality. So, we could improve reusability making all (less attributes) public. Today, I can see no funcionality for these classes, but in the future? If there is no funcionality, the user won't use them. What do you think?

Timmeey: Could you recheck my comment? Unfortunately i edited it while your were already answering it

fabriciofx: @Timmeey You're proposing something like that?
```java
public CapitalizedText(final Text text) {
   this(
      text,
      new Func<Text, Text>() {
          @Override
          public Text apply(final Text input) throws Exception {
             return new FormattedText(
               "%s%s",
               input.asString().substring(0, 1).toUpperCase(),
               input.asString().substring(1)
            );
         }         
      };
   );
}
public CapitalizedText(final Text text, final Func<Text, Text> func) {
  this.origin = text;
  this.func = func;
}
```
If yes, I liked this proposition. 😄

Timmeey: I didn't have the secondary constructor in mind, but if you do it that way,  the primary one would need to be private.
Anyway, why the second ctor? you would never want to give it a different capitalize function?! Just keep the first ctor and remove the second one.

And if we could replace this hideous java extra verbosity with lambdas
```
public class CapitalizedText implements Text {
    private final Text origin;
    private final Func<Text, Text> func;

    public CapitalizedText(final Text text) {
        this.origin = text;
        this.func = input -> new FormattedText(
            "%s%s",
            input.asString().substring(0, 1).toUpperCase(),
            input.asString().substring(1)
        );
   
    }
```
i would be perfectly satisfied without anything left to complain

fabriciofx: @Timmeey Primary one constructor as private? With all due to respect, but no way! 😄 In general, I don't like any code in constructors. I know is ok, because it's not executing, just creating a new object, but I think it not elegant. So, we have these options:
1. To declare as a public class, as it's now;
2. To declare the function as anonymous, as it was before;
3. To declare in the secundary constructor, as I suggested;
4. To declare in the primary constructor, as you suggested;
5. To declare as a private class.

So, what do you prefer?
@yegor256 Would do you like give us your opinion?

Timmeey: Or 6. Option
Declare the Capitalize Function as a "static" (not static as the word static) Field of the Class

```java
public class CapitalizedText implements Text {
    private final Text origin;
    private final Func<Text, Text> func = input -> new FormattedText(
            "%s%s",
            input.asString().substring(0, 1).toUpperCase(),
            input.asString().substring(1)
        );

    public CapitalizedText(final Text text) {
        this.origin = text;
    }
```

I just don't see how an extra constructor would improve the situation.

But are we in agreement that the Function looks nicer as a lambda statement?

src/main/java/org/cactoos/text/JoinedText.java
```diff
@@ -0,0 +1,86 @@
...
+    public JoinedText(final Iterable<Text> texts) {
+        this(" ", texts);
+    }
+
+    /**
+     * Ctor.
+     * @param delimiter Delimit among texts
+     * @param texts Texts to be joined
+     */
+    public JoinedText(final String delimiter, final Iterable<Text> texts) {
```
Timmeey: Should we try to also use Text as the type of the delimiter?

fabriciofx: @Timmeey I changed the delimiter and create a new secondary constructor. WDYT?

fabriciofx: @Timmeey Thanks for your review!

src/main/java/org/cactoos/text/CapitalizeWordFunc.java
```diff
@@ -0,0 +1,48 @@
...
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.3
+ */
+public final class CapitalizeWordFunc implements Func<Text, Text> {
+    @Override
+    public Text apply(final Text input) throws Exception {
+        return new FormattedText(
```
Timmeey: this is my last remark. i promise...


But since we have Subtext, we should use it here instead of the buildin static method.
```java
return new JoinedText(
  new UpperCaseText(
     new Subtext(input,0,1)),
  new Subtext(input,1)
);
```

fabriciofx: @Timmeey Please, feel free to comment how many times you want! Your reviews and comments are amazing and I'm learning a lot of with them! I saw the same problem but before to correct it I'm thinking how can I implement the `SubText(Text, int)` constructor, because to do it I'll need put a `text.asString().length()` in the constructor, what is inacceptable. What do you think?

Timmeey: @fabriciofx I'm flattered. Thanks.

you are right `text.asString().length()` in a Ctor is unacceptable. Should we just mimic the behaviour of the JDK that when we don't supply a endIndex, we just take everything util the end? 

If NOT.... I have an idea :-P
Instead of using SubText, we could introduce new HeadText(Text input, int endIndex), that gives the full text until the endIndex (Head of the Text), and a new TailText(Text input, int startIndex) which gives you everything from the startIndex until the end. Then there would be no confusion about how to signal you want subtext to return everything until the end.

```java
return new JoinedText(
  new UpperCaseText(
     new HeadText(input,0)),
  new TailText(input,1)
);
```

@yegor256 , WDYT about introducing Head/Tail-Text to avoid using "Flags" in SubText to signal intent to go until the end/start of the word?

yegor256:
@fabriciofx this PR is too big. I would suggest to break it down to smaller parts.

src/main/java/org/cactoos/text/CapitalizeWordFunc.java
fabriciofx: @Timmeey Your idea is very interesting, but we still have the problem to select a subtext in the middle of a `Text`. To solve it, I'm thinking in introduce an `LengthOfText` class, to return the length of a `Text`.

fabriciofx:
@yegor256 Ok. I'll do it. Thanks for your attention!

src/main/java/org/cactoos/text/RepeatedText.java
```diff
@@ -0,0 +1,68 @@
...
+     * @param count How many times repeat the Text
+     */
+    public RepeatedText(final Text text, final int count) {
+        this.origin = text;
+        this.count = count;
+    }
+
+    @Override
+    public String asString() throws IOException {
+        final StringBuilder out = new StringBuilder(0);
```
fabriciofx: @englishman I was just following Yegor... he did something like it in some code... 😄 But I agree. I'll change it. Thanks for your attention!

src/main/java/org/cactoos/text/SplitText.java
```diff
@@ -0,0 +1,103 @@
...
+    public Iterator<Text> iterator() {
+        try {
+            return new TransformedIterable<>(
+                new ArrayAsIterable<>(
+                    this.origin.asString()
+                    .split(this.pattern, this.limit)
+                ),
+                new StringAsTextFunc()
+            ).iterator();
+        } catch (final IOException ex) {
```
fabriciofx: @englishman I think so. `this.origin.asString()` throws `IOException`. WDYT?

fabriciofx:
@yegor256 I've a question: I'm spliting this PR in some smaller PRs, as you requestered, but the problem is some classes have other classes in the same PR as dependency. So, the PR can be stay big. What do you recommend? Thanks in advance!

yegor256:
@fabriciofx start with something that everything else depends on.

fabriciofx:
Closed this PR because the commits were broken in others.

Issue #74, opened 2017-06-02
UncheckedScalar


yegor256:
Let's create `UncheckedScalar`, `UncheckedText`, and `UncheckedBytes`. Similar to `UncheckedFunc`

yegor256:
I believe it's fixed in #81

fabriciofx:
@yegor256 Yeah, could you close this issue?

yegor256:
@fabriciofx thanks!

PR #81, opened 2017-06-03
#74: UncheckedScalar, UncheckedText, and UncheckedBytes

fabriciofx:


0crat:

src/main/java/org/cactoos/func/UncheckedScalar.java
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Scalar that doesn't throw checked {@link Exception}.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @param <T> Type of result
+ * @since 0.2
```
fabriciofx: @g4s8 Ops, my mistake. I'll change it. Thanks!

src/main/java/org/cactoos/func/UncheckedScalar.java
```diff
@@ -0,0 +1,63 @@
...
+    public UncheckedScalar(final Scalar<T> scalar) {
+        this.scalar = scalar;
+    }
+
+    @Override
+    public T asValue() {
+        try {
+            return this.scalar.asValue();
+        } catch (final IOException ex) {
+            throw new IllegalStateException(ex);
```
fabriciofx: @g4s8 Well, I just followed the same approach was used [here](https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/func/UncheckedFunc.java). 😄 
But you're right. `IllegalStateException` isn't the better exception in this case. Fell free to suggest another one.

g4s8: @fabriciofx maybe `java.io.UncheckedIOException` would be better? https://docs.oracle.com/javase/8/docs/api/java/io/UncheckedIOException.html

fabriciofx: @g4s8 I think it very good. Thanks!

yegor256:
@rultor try to merge

PR #84, opened 2017-06-05
#53 IterableAsList needs cached size()

### LINKED ISSUE #53: IterableAsList.java:56-58: Needs cached `LengthOfIterable`...
The puzzle `39-e7c34bd5` in [`src/main/java/org/cactoos/list/IterableAsList.java`](https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/list/IterableAsList.java) (lines 56-58) has to be resolved: "Needs cached `LengthOfIterable` version to improve `IterableAsList` performance. Now each call to `size()` goes through all iterable to calculate the size."

The puzzle was created by g4s8 on 27-May-17. 

Estimate: 30 minutes, role: IMP.

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and
[about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

Timmeey:
This PR is based on PR #81 (would need to be merged prior to this PR) due to the usage of UncheckedScalar to be able to use CachedScalar class in a safe way (without requiring to declare size() an unsafe operation).
Closes #53 

Reasoning about the placement of the cache into IterableAsList and not into LengthOfIterable/LengthOfIterator

- LengthOfIterator: is a one-time use Object, it consumes the Iterator, therefore caching would alter how this works
- LengthOfIterable: Seemed a good spot, but the underlying iterable could change, therefore making the cached value inconsistent
- IterableAsList: Same argument here, the underlying iterable could change, but if it would change while being used here, the usage of this Class would not make much sense. Therefore the correct place to use the caching mechanism is in this class directly

### DIFF
--- a/src/main/java/org/cactoos/func/CachedScalar.java
+++ b/src/main/java/org/cactoos/func/CachedScalar.java
@@ -1,0 +1,69 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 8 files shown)


0crat:
@timmeey I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

yegor256:
@rultor try to merge

Issue #86, opened 2017-06-06
FuncWithCallback with three arguments


yegor256:
Let's add one more optional argument to `FuncWithCallback`, which will be applied if no exception is thrown.

yegor256:
done in #87

yegor256:
@rultor release, tag is `0.3`

PR #87, opened 2017-06-06
86: third arg for FuncWithCallback

yegor256:
as per #86: third argument for `FuncWithCallback` ctor introduced


yegor256:
@rultor try to merge

PR #91, opened 2017-06-08
Collection as text

memoyil:
Collection as text

### DIFF
--- a/src/main/java/org/cactoos/text/CollectionAsText.java
+++ b/src/main/java/org/cactoos/text/CollectionAsText.java
@@ -1,0 +1,67 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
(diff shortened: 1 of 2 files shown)


0crat:
@memoyil I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

src/main/java/org/cactoos/text/CollectionAsText.java
```diff
@@ -0,0 +1,67 @@
...
+     * The delimeter.
+     */
+    private final CharSequence delimeter;
+
+    /**
+     * Ctor.
+     * @param collection The Collection
+     * @param seperator The seperator
+     */
+    public CollectionAsText(final Collection<? extends String> collection, final
```
Timmeey: String is final, therefore `<? extends String>` is redundant

Timmeey:
Take a look at #80 https://github.com/yegor256/cactoos/pull/80/files#diff-0e5a852a570d9dcfd2b2c143ee8f2d17

```
Collection<String> stringsCollection=......
new JoinedText(
  new TransformedIterable(
    stringsCollection,
    new StringAsTextFunc())
```
will exactly do what you propose.
So if you think that should be a buildIn operation, try to compose this witht eh already present classes like the way i sceteched it above (the PR is not merged yet, but it will be soon)

memoyil:
@Timmeey I think JoinedText and CollectionAsText are different issues and both of them should be inside the API.

Timmeey:
But isn't that exactly what you do?
You put in a Collection of Strings (which can be seen as an iterable, since COllection implements iterable)
And you get out one single long text, join on your charSequence Delimiter.

JoinedText takes a collection/iterable of Texts, 
and puts out a long text, joined on the delimiter.

Sorry, i don't really see the difference here.

Maybe give an example of how you would use them differently?

memoyil:
@Timmeey  Okay, that's correct. I accept your critics. Let's get another comments from other contributors and Yegor also. I dont say that mine's correct and we should apply this PR. I think so as you also think like that.

yegor256:
@memoyil this one seems redundant to me if we have `JoinedText`

memoyil:
@yegor256 Ok i am closing the pull request

src/main/java/org/cactoos/text/CollectionAsText.java
```diff
@@ -0,0 +1,67 @@
...
+     */
+    private final CharSequence delimeter;
+
+    /**
+     * Ctor.
+     * @param collection The Collection
+     * @param seperator The seperator
+     */
+    public CollectionAsText(final Collection<String> collection, final
+        CharSequence seperator) {
```
englishman: Can we put this on prev. line?

src/main/java/org/cactoos/text/CollectionAsText.java
```diff
@@ -0,0 +1,67 @@
...
+        this.source = collection;
+        this.delimeter = seperator;
+    }
+
+    @Override
+    public String asString() {
+        return this.source.stream()
+            .map(i -> i.toString())
```
englishman: Don't needed at all as `source` defined already as `Collection<String> source`

src/test/java/org/cactoos/text/CollectionAsTextTest.java
```diff
@@ -0,0 +1,53 @@
...
+ * @checkstyle JavadocMethodCheck (500 lines)
+ */
+public final class CollectionAsTextTest {
+
+    @Test
+    public void collectionAsText() {
+        MatcherAssert.assertThat(
+            "Can't contruct a text from collection",
+            new CollectionAsText(
+                new ArrayList<String>(Arrays.asList("Hello", "World")), ","
```
englishman: `new ArrayList<String>(...)` wrapper isn't needed here

Issue #98, opened 2017-06-09
Repeat must be broken into two parts


yegor256:
`Repeat` design is not perfect

yegor256:
fixed in #99

PR #99, opened 2017-06-09
98: EndlessIterator

yegor256:
as per #98:

  * `EndlessIterator` introduced
  * `EndlessIterable` introduced
  * `Repeat` renamed to `RepeatIterable`
  * `RepeatIterator` is now public


yegor256:
@rultor merge

PR #108, opened 2017-06-11
First or default

g4s8:
First element in iterable with fallback value if iterable is empty.


0crat:

yegor256:
@g4s8 good idea, but I would add an optional param to `FirstOf`. We have a similar design in `FuncWithFallback`. I don't like two classes, which are doing very similar things.

g4s8:
@yegor256 I see. Moved fallback to `FIrstOf` contructor param.

src/main/java/org/cactoos/list/FirstOf.java
```diff
@@ -40,23 +42,57 @@
...
     /**
      * Ctor.
      *
-     * @param source Iterable
+     * @param src Iterable
+     * @param fbk Fallback value
      */
-    public FirstOf(final Iterable<T> source) {
-        this.source = source;
+    public FirstOf(final Iterable<T> src, final Scalar<T> fbk) {
```
yegor256: @g4s8 looks good, but I feel that we're not using `Scalar` the right way here. We're losing the context. This scalar won't have any idea of why it's called and what is the situation there. I would rather use `Func<Iterable<T>, T>`. This is how that `Func` will be able to throw a more meaningful exception.

yegor256:
@g4s8 see my comment above

g4s8:
@yegor256 you're right, replaced `Scalar<T>` with `Func<Iterable<T>, T>`.

yegor256:
@rultor merge

Issue #109, opened 2017-06-12
LimitedIterable


yegor256:
Let's add `LimitedIterable` similar to `Iterables.limit()` of Guava

dusan-rychnovsky:
I am working on this one. I'll send a PR soon.

dusan-rychnovsky:
fixed in #110

yegor256:
@dusan-rychnovsky thanks!

PR #110, opened 2017-06-12
#109 LimitedIterable

dusan-rychnovsky:
Add LimitedIterable - an analogy Iterables.limit from Guava.

### DIFF
--- a/src/main/java/org/cactoos/list/LimitedIterable.java
+++ b/src/main/java/org/cactoos/list/LimitedIterable.java
@@ -1,0 +1,40 @@
+package org.cactoos.list;
+
+import java.util.Iterator;
+
+/**
+ * Limited iterable.
+ *
+ * <p>This is a view of an existing iterable containing the given number of its
+ * first elements.</p>
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @param <T> Element type
+ * @since 0.4
+ */
+public class LimitedIterable<T> implements Iterable<T> {
+  
+    private final Iterable<T> iterable;
+    private final int limitSize;
+
+    /**
+     * Ctor.
+     *
+     * @param iterable the underlying iterable
+     * @param limitSize the requested number of elements
+     */
+    public LimitedIterable(Iterable<T> iterable, int limitSize) {
+        if (limitSize < 0) {
+            throw new IllegalArgumentException(
+                "Expected limitSize >= 0, got [" + limitSize + "]."
+            );
+        }
+        this.iterable = iterable;
+        this.limitSize = limitSize;
+    }
+
+    @Override
+    public Iterator<T> iterator() {
+        return new LimitedIterator<>(iterable.iterator(), limitSize);
+    }
+}
--- a/src/main/java/org/cactoos/list/LimitedIterator.java
+++ b/src/main/java/org/cactoos/list/LimitedIterator.java
@@ -1,0 +1,49 @@
+package org.cactoos.list;
+
+import java.util.Iterator;
+
+/**
+ * Limited iterator.
+ *
+ * <p>This is a decorator over an existing iterator. Returns elements of the
+ * original iterator, until either the requested number of items have been
+ * returned or the underlying iterator has been exhausted.</p>
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @param <T> Element type
+ * @since 0.4
+ */
+public class LimitedIterator<T> implements Iterator<T> {
+  
+    private final Iterator<T> iterator;
+    private final int limitSize;
+    private int consumedSize;
+
+    /**
+     * Ctor.
+     *
+     * @param iterator the underlying iterator
+     * @param limitSize the requested number of elements
+     */
+    public LimitedIterator(Iterator<T> iterator, int limitSize) {
+        if (limitSize < 0) {
+            throw new IllegalArgumentException(
+                "Expected limitSize >= 0, got [" + limitSize + "]."
+            );
+        }
+        this.iterator = iterator;
+        this.limitSize = limitSize;
+        this.consumedSize = 0;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return (consumedSize < limitSize) && iterator.hasNext();
+    }
+
+    @Override
+    public T next() {
+        ++consumedSize;
+        return iterator.next();
+    }
+}
--- a/src/test/java/org/cactoos/list/LimitedIterableTest.java
+++ b/src/test/java/org/cactoos/list/LimitedIterableTest.java
@@ -1,0 +1,62 @@
+package org.cactoos.list;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+/**
+ * Test case for {@link LimitedIterable}
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @since 0.4
+ */
+public class LimitedIterableTest {
+
+    @Test
+    public void iteratesOverPrefixOfGivenLength() {
+        MatcherAssert.assertThat(
+            "Can't limit an iterable with more items",
+            new LimitedIterable<>(
+                new ArrayAsIterable<>(0, 1, 2, 3, 4),
+                3
+            ),
+            Matchers.contains(0, 1, 2)
+        );
+    }
+
+    @Test
+    public void iteratesOverWholeIterableIfThereAreNotEnoughItems() {
+        MatcherAssert.assertThat(
+            "Can't limit an iterable with less items",
+            new LimitedIterable<>(
+                new ArrayAsIterable<>(0, 1, 2, 3, 4),
+                10
+            ),
+            Matchers.contains(0, 1, 2, 3, 4)
+        );
+    }
+  
+    @Test
+    public void limitOfZeroProducesEmptyIterable() {
+        MatcherAssert.assertThat(
+            "Can't limit an iterable to zero items",
+            new LimitedIterable<>(
+                new ArrayAsIterable<>(0, 1, 2, 3, 4),
+                0
+            ),
+            Matchers.iterableWithSize(0)
+        );
+    }
+
+    @Test
+    public void emptyIterableProducesEmptyIterable() {
+        MatcherAssert.assertThat(
+            "Can't limit an empty iterable",
+            new LimitedIterable<>(
+                new ArrayAsIterable<>(),
+                10
+            ),
+            Matchers.iterableWithSize(0)
+        );
+    }
+}


0crat:
@dusan-rychnovsky I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

src/main/java/org/cactoos/list/LimitedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+ * Limited iterable.
+ *
+ * <p>This is a view of an existing iterable containing the given number of its
+ * first elements.</p>
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @param <T> Element type
+ * @since 0.4
+ */
+public class LimitedIterable<T> implements Iterable<T> {
```
Timmeey: Classes are either abstract or final. We should make it final

src/main/java/org/cactoos/list/LimitedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+    private final int limitSize;
+
+    /**
+     * Ctor.
+     *
+     * @param iterable the underlying iterable
+     * @param limitSize the requested number of elements
+     */
+    public LimitedIterable(Iterable<T> iterable, int limitSize) {
+        if (limitSize < 0) {
```
Timmeey: No code in Ctors.
If someone supplies something <0 it should work fine, the underlying iterator would just not return anything

src/main/java/org/cactoos/list/LimitedIterator.java
```diff
@@ -0,0 +1,49 @@
...
+    private int consumedSize;
+
+    /**
+     * Ctor.
+     *
+     * @param iterator the underlying iterator
+     * @param limitSize the requested number of elements
+     */
+    public LimitedIterator(Iterator<T> iterator, int limitSize) {
+        if (limitSize < 0) {
```
Timmeey: No code in Ctors. If someone supplies something <0 hasNext() will return false.

src/main/java/org/cactoos/list/LimitedIterator.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @Override
+    public boolean hasNext() {
+        return (consumedSize < limitSize) && iterator.hasNext();
+    }
+
+    @Override
+    public T next() {
+        ++consumedSize;
```
Timmeey: call hasNext() prior to doing anything in here. if hasNext() returns false, throw NoMoreElementsException.

src/main/java/org/cactoos/list/LimitedIterator.java
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * <p>This is a decorator over an existing iterator. Returns elements of the
+ * original iterator, until either the requested number of items have been
+ * returned or the underlying iterator has been exhausted.</p>
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @param <T> Element type
+ * @since 0.4
+ */
+public class LimitedIterator<T> implements Iterator<T> {
```
Timmeey: Classes are either abstract or final. We should make this final

src/main/java/org/cactoos/list/LimitedIterator.java
```diff
@@ -0,0 +1,49 @@
+package org.cactoos.list;
+
+import java.util.Iterator;
+
+/**
+ * Limited iterator.
+ *
```
Timmeey: Note that this gives no thread safety guarantees.

(Concurrent calls to next(), will lead to overshooting the limit)

yegor256:
@dusan-rychnovsky I agree with all comments made by @Timmeey above

src/main/java/org/cactoos/list/LimitedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+    private final Iterable<T> iterable;
+    private final int limitSize;
+
+    /**
+     * Ctor.
+     *
+     * @param iterable the underlying iterable
+     * @param limitSize the requested number of elements
+     */
+    public LimitedIterable(Iterable<T> iterable, int limitSize) {
```
yegor256: @dusan-rychnovsky it's better to use a different name here, in order to avoid collisions. Also, pay attention that compound names are a bad practice: http://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html

src/main/java/org/cactoos/list/LimitedIterable.java
```diff
@@ -0,0 +1,40 @@
...
+        if (limitSize < 0) {
+            throw new IllegalArgumentException(
+                "Expected limitSize >= 0, got [" + limitSize + "]."
```
yegor256: @dusan-rychnovsky string concatenation is a bad practice: http://www.yegor256.com/2014/06/19/avoid-string-concatenation.html

src/main/java/org/cactoos/list/LimitedIterator.java
```diff
@@ -0,0 +1,49 @@
...
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @param <T> Element type
+ * @since 0.4
+ */
+public class LimitedIterator<T> implements Iterator<T> {
+  
+    private final Iterator<T> iterator;
+    private final int limitSize;
+    private int consumedSize;
```
yegor256: @dusan-rychnovsky avoid compound names, they are evil

src/main/java/org/cactoos/list/LimitedIterator.java
yegor256: @dusan-rychnovsky check this too: http://www.yegor256.com/2015/04/30/iterating-adapter.html

src/test/java/org/cactoos/list/LimitedIterableTest.java
```diff
@@ -0,0 +1,67 @@
...
+ * Test case for {@link LimitedIterable}
+ *
+ * @author Dusan Rychnovsky (dusan.rychnovsky@gmail.com)
+ * @since 0.4
+ */
+public class LimitedIterableTest {
+
+    @Test(expected = IllegalArgumentException.class)
+    public void failsIfGivenLimitIsLessThanZero() {
+       new LimitedIterable<>(new ArrayAsIterable<>(), -1);
```
yegor256: @dusan-rychnovsky I don't think there is anything wrong with the negative number here. Did you check what Guava is doing in this case?

yegor256:
@dusan-rychnovsky there are some comments from me too

src/test/java/org/cactoos/list/LimitedIterableTest.java
dusan-rychnovsky: Hi @yegor256, 

From the documetation of Guava - `throws IllegalArgumentException if limitSize is negative` (https://google.github.io/guava/releases/21.0/api/docs/com/google/common/collect/Iterables.html#limit-java.lang.Iterable-int-).

I believe this is the correct approach. If someone is trying to get a negative number of elements from an iterable, there is most probably a bug in the code somewhere before. We want to fail fast. Not swallow the error and return some pseudo-correct result.

WDYT?

dusan-rychnovsky: cc @Timmeey

Timmeey:
@dusan-rychnovsky I concur. 
As long as we don't throw the exception in the ctor. 

Am 13. Juni 2017 08:04:43 MESZ schrieb "Dušan Rychnovský" <notifications@github.com>:

src/test/java/org/cactoos/list/LimitedIterableTest.java
dusan-rychnovsky: I don't see how this behavior can be achieved without throwing an exception from inside the constructor. This is no real code in ctor - rather this is an assertion.

yegor256:
@dusan-rychnovsky I agree, let's throw if the number is negative. But I agree with @Timmeey too, we must not do that in the ctor.

dusan-rychnovsky:
@Timmeey @yegor256 

I believe we need a way to prevent instantiation inside object's constructor when the outside world provides invalid input parameters and the object cannot be correctly instantiated. What's the point of having invalid objects around?

Throwing the exception in `iterator` method is much worse. This is not fail fast. The error would surface much later than when we first got the chance to find about it. This will hide the error and make it more difficult to debug.

And for `LimitedIterator` it's not even clear where else to throw the exception instead of in constructor, because there are multiple methods - hasNext and next.

Assertions / PreConditions should be placed to the top of the method which accepts the parameter.

yegor256:
@dusan-rychnovsky we're trying to follow this philosophy: http://www.yegor256.com/2015/05/07/ctors-must-be-code-free.html Validating means work and it must not be done while objects are not yet ready and don't have control. In this particular case it's just `int`. But it that would be something like `Number` connected to the database -- you would see the problem:

```
new LimitedIterable(origin, new MySQLCounter("SELECT COUNT(n) FROM x"));
```

See? In order to calculate the number, we should do a round-trip to the MySQL server. This is *not* what you want to happen when the object is not ready yet.

dusan-rychnovsky:
@yegor256

If you need to query the limit from a database then this needs to be done on the caller side and provided to `LimitedIterable` via constructor. From the `LimitIterable`'s perspective, it will always be "just an int". It is the callers responsibility to validate the limit.

This exception does not serve to validate user input, it is not even intended to be caught / handled. It is an assertion and it's only purpose is to catch errors in code. The reason why I'm using an explicit if stmt + exception is that Java lacks a reasonable `assert` construct.

If you decide to condemn assertions, in my opinion you're loosing a valuable quality-assurance tool.

Anyway, I don't see how this check can be done other than in constructor, so I'm going to omit it completely.

src/main/java/org/cactoos/list/LimitedIterable.java
dusan-rychnovsky: @yegor256 I understand the second part, but which name are you referring to regarding avoiding collisions - the decorated iterable? Please explain. Thanks

src/main/java/org/cactoos/list/LimitedIterable.java
dusan-rychnovsky: Fixed

src/main/java/org/cactoos/list/LimitedIterable.java
dusan-rychnovsky: Removed

src/main/java/org/cactoos/list/LimitedIterator.java
dusan-rychnovsky: Mentioned this in javadoc

src/main/java/org/cactoos/list/LimitedIterator.java
dusan-rychnovsky: Fixed

src/main/java/org/cactoos/list/LimitedIterator.java
dusan-rychnovsky: Removed

src/main/java/org/cactoos/list/LimitedIterator.java
dusan-rychnovsky: Fixed

dusan-rychnovsky:
One more question - how does this `@version` annotation works? It looks like it contains some random hash. How do you generate the hash? Thanks

yegor256:
@dusan-rychnovsky check this out: http://www.yegor256.com/2017/05/10/inversion-of-control.html We don't want our clients to pass any params to us. Instead, we want them to give us objects and we'll pull data out of them when needed.

yegor256:
@dusan-rychnovsky just copy that `@version` lines from other files

dusan-rychnovsky:
@yegor256 I believe this PR can be merged. Thanks

dusan-rychnovsky:
@rultor merge

dusan-rychnovsky:
Hmm does not seem to work for me :)

yegor256:
@rultor try to merge

Issue #144, opened 2017-06-15
Predicate


VsSekorin:
Why was Pred (Predicate, https://github.com/yegor256/cactoos/pull/35) removed? (https://github.com/yegor256/cactoos/pull/72/commits/f1fe71b7bb3f884b8c69aaa930bfb34659a19f54)
Let's restore it (https://github.com/yegor256/cactoos/issues/36).

0crat:
@vssekorin I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

yegor256:
@VsSekorin right, let's add it as `org.cactoos.Pred`

PR #145, opened 2017-06-15
#144 Predicate

VsSekorin:
As per https://github.com/yegor256/cactoos/issues/144


0crat:

yegor256:
@VsSekorin this one doesn't make sense since it's exactly the same as `Func<T, Boolean>`. I think that we need it to return boolean and accept no arguments.

VsSekorin:
@yegor256 no arguments. If I understand correctly, It will be logical value, not smth like java.util.function.Predicate?

VsSekorin:
`[ERROR] /home/travis/build/yegor256/cactoos/src/test/java/org/cactoos/io/StickyInputTest.java:[46,33] incompatible types: cannot infer type-variable(s) T,X,Y`

**StickyInputTest** I don't know anything about this.

VsSekorin:
**StickyInputTest** Because of this, the last PR were not build. I don't know anything about this class.

yegor256:
@VsSekorin maybe `Term` would be a better name? https://en.wikipedia.org/wiki/Term_(logic)

VsSekorin:
@yegor256 The term is a broader concept. Renamed to `LogTerm`.

src/main/java/org/cactoos/LogTerm.java
```diff
@@ -0,0 +1,44 @@
...
+package org.cactoos;
+
+/**
+ * Logical value.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Vseslav Sekorin (vssekorin@gmail.com)
+ * @version $Id$
+ * @since 0.5
```
yegor256: @VsSekorin it's 0.7

yegor256:
@VsSekorin and now it's exactly the same as `Scalar<Boolean>` :) I don't see why we need this new interface if it just duplicates what we already have. WDYT?

VsSekorin:
@yegor256 my bad. The duplicate interface was deleted.

VsSekorin:
@yegor256 I'm not sure, but it looks like the problem with Codecov is not in this PR.

yegor256:
@rultor merge

Issue #151, opened 2017-06-16
Web url based resource links bring to test failure


englishman:
If network environment is restricted or development is going on offline then some tests are failed.
E.g. such links
```
new UrlAsInput(
    "https://raw.githubusercontent.com/yegor256/cactoos/0.5/pom.xml"
)
```
have to be changed to local files with `"file:///<relative path here>/pom.xml"`

0crat:
@englishman I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

yegor256:
@englishman indeed it's a bug, thanks.

englishman:
Will take care of it

yegor256:
@englishman I would not change them to `file://...` It's important to have real tests that actually go to real URLs. In order to see how we behave with content that is not available immediately, but takes some time to read. I would rather add some gate keepers in front of the tests, which will check whether we're online now or not. But I'm not sure... :)

englishman:
Set of broken tests
![6-16-2017 11-04-06 am](https://user-images.githubusercontent.com/1883572/27232348-bde8e8f2-5283-11e7-8eda-ae36158acca5.png)

englishman:
@yegor256 
> It's important to have real tests that actually go to real URLs.

Even with `file://...` it'll go through real `URL` java class. And we don't want to test it behavior.

> In order to see how we behave with content that is not available immediately, but takes some time to read.

Honestly, I can't see any reason for this testing primitive utility classes. But if you see, please, explain.

> But I'm not sure... :)

Will provide `file://...` bases fix and if we really need slow `org.cactoos.Input` implementation we'll create it for such testing purposes later.

englishman:
Fixed

PR #164, opened 2017-06-16
change external test resource to file system file

englishman:
Fix tests for #151

### DIFF
--- a/src/test/java/org/cactoos/io/LengthOfInputTest.java
+++ b/src/test/java/org/cactoos/io/LengthOfInputTest.java
@@ -75,11 +75,11 @@ public final class LengthOfInputTest {
             new LengthOfInput(
                 new UrlAsInput(
                     // @checkstyle LineLength (1 line)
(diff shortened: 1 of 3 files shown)


0crat:
@englishman I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

fabriciofx:
@englishman I think its better use Takes to running a https server and provide a file to be retrieved, similar to [here](https://github.com/yegor256/cactoos/blob/master/src/test/java/org/cactoos/io/UrlAsInputTest.java#L59-L77). WDYT?

englishman:
@fabriciofx I don't think it's an option as `FtRemote.Script` impl gives to test code `URI` but problem situation requires `String` and `URI` instance is created from string in a classes under the test.

Honestly, I don't understand why Take framework was used even in your example as it looks like an attempt to test also an JDK `URL` class we actually don't need.

yegor256:
@rultor merge

Issue #172, opened 2017-06-18
Max and Min


fabriciofx:
Let's implement Max and Min. Max returns the greater and Min returns smaller among numbers.

0crat:
@fabriciofx I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

fabriciofx:
Closed #263

PR #173, opened 2017-06-18
#172: Max and Min

fabriciofx:
As per #172:

- `Max` introduced.
- `Min` introduced.


0crat:

englishman:
I guess new ctors should be added here for consuming `Scalar<Number>` to allow combination of `Min` and `Max`

fabriciofx:
@englishman I saw it, but I think the best way is change `Max` and `Min` to extends `Number`.  WDYT?

src/main/java/org/cactoos/math/Max.java
```diff
@@ -0,0 +1,109 @@
...
+ * Find the greater among numbers.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.7
+ */
+@SuppressWarnings("serial")
+public final class Max extends Number implements Scalar<Number> {
```
yegor256: @fabriciofx I would rather do:

```java
class Max<T extends Comparable<T>> implements Scalar<T> {
}
```

And then use `compareTo()` in its body.

src/main/java/org/cactoos/math/Max.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Numbers.
+     */
+    private final Number[] numbers;
```
yegor256: @fabriciofx I think it's better to encapsulate `Iterable<Scalar<T>>`

yegor256:
@fabriciofx good idea, but I would do it differently, see my comments

src/main/java/org/cactoos/math/Max.java
englishman: @yegor256 it would be more logical to use `<T extends Number>` then `<T extends Comparable<T>>` as `new Max("string 1", "string 2", "string 3").asValue()` looks illogical, IMHO.

fabriciofx:
@yegor256 I got it. Thanks!

src/main/java/org/cactoos/math/Max.java
fabriciofx: @englishman I think the @yegor256 idea is make `Max` and `Min` support any thing, not just numbers. I'm working in a change, and will appreciate your comments on the new commit. Thanks for your attention!

fabriciofx:
@yegor256 Could you check it now?

src/main/java/org/cactoos/math/Max.java
fabriciofx: @yegor256 Done. Thanks.

yegor256:
@fabriciofx again the name of the branch will cause problems in Rultor, please rename

fabriciofx:
@yegor256 This PR was renamed to #263 .

PR #174, opened 2017-06-18
ComposedFunc

VsSekorin:
https://github.com/yegor256/cactoos/pull/15 reincarnation


0crat:

src/main/java/org/cactoos/func/ComposedFunc.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Composed function.
+ *
+ * @author Vseslav Sekorin (vssekorin@gmail.com)
+ * @version $Id$
+ * @param <X> Type of input.
+ * @param <Y> Type of output.
+ * @since 0.7
+ */
+public final class ComposedFunc<X, Y> implements Func<X, Y> {
```
yegor256: @VsSekorin why not `ComposedFunc<X, Y, T>`?

src/main/java/org/cactoos/func/ComposedFunc.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param iterator Functions
+     * @param input The input
+     * @return Functions apply
+     * @throws Exception If fails
+     */
+    private Object funcsApply(
+        final Iterator<Func<Object, Object>> iterator,
+        final Object input
+    ) throws Exception {
+        // @checkstyle AvoidInlineConditionalsCheck (3 lines)
```
yegor256: @VsSekorin just don't use it, no need to suppress the checkstyle

src/test/java/org/cactoos/func/ComposedFuncTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Test
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
+    public void withoutIterable() throws Exception {
+        final Integer count = new LengthOfIterable(
+            new FilteredIterable<>(
+                new TransformedIterable<>(
+                    new ArrayAsIterable<>("public", "final", "class"),
+                    new ComposedFunc<Object, Object>(
+                        input -> ((String) input).concat("1"),
```
yegor256: @VsSekorin type casting is evil, don't use it anywhere, ever: http://www.yegor256.com/2015/04/02/class-casting-is-anti-pattern.html

yegor256:
@VsSekorin see my comments above

VsSekorin:
@yegor256 I completely agree with all the comments. These problems are caused by `Iterator<Func<Object, Object>>`. I propose to go back to `ComposedFunc<X, Y, Z>`, as it was in #15

yegor256:
@VsSekorin yes, that's how it should be: `ComposedFunc<X, Y, Z>`

VsSekorin:
@yegor256 I thought about the option with 
`Iterable<Func<X, X>>` => `Func<X, Y>` => `Iterable<Func<Y, Y>>` => `Func<Y, Z>` => `Iterable<Func<Z, Z>>`
, but more than 3 parameters in ctor. And class would be too big because of 8 ctor.

src/main/java/org/cactoos/func/ComposedFunc.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Composed functions in this.func.
+     *
+     * @param iterator Functions
+     * @param input The input
+     * @return Functions apply
+     * @throws Exception If fails
+     */
+    private Y funcsApply(
```
yegor256: @VsSekorin in general it's not a good idea to have private methods in a class. Try to avoid them.

yegor256: @VsSekorin also, I think you're abusing the recursion. It makes code more complicated and less readable. In this particular case it would much easier to use plain old `for` loop.

src/test/java/org/cactoos/func/ComposedFuncTest.java
```diff
@@ -0,0 +1,90 @@
...
+                    new ComposedFunc<>(
+                        input -> input.concat("1"),
+                        input -> input.concat("2")
+                    )
+                ),
+                input -> input.endsWith("12")
+            )
+        ).asValue();
+        MatcherAssert.assertThat(
+            count,
```
yegor256: @VsSekorin why do you need this variable? Can't you just use value here?

src/test/java/org/cactoos/func/ComposedFuncTest.java
```diff
@@ -0,0 +1,90 @@
...
+                            input -> input.replaceAll("a", "b")
+                        ),
+                        String::trim
+                    )
+                ),
+                input -> !input.startsWith("st")
+            )
+        ).asValue();
+        MatcherAssert.assertThat(
+            count,
```
yegor256: @VsSekorin same here, why variable?

yegor256:
@VsSekorin see my comments above. Also, I don't really like the name `ComposedFunc`. Maybe something like `TrainFunc` or `ChainedFunc` or `ChainOfFuncs`?

VsSekorin:
@yegor256 Made the necessary corrections.

yegor256:
@rultor merge

VsSekorin:
@yegor256 Try now, please.

yegor256:
@rultor merge

Issue #180, opened 2017-06-19
unnecessary immutability


yegor256:
I believe we made a mistake in:

```
ArrayAsIterable
IterableAsList
IterableAsMap
MapAsProperties
```

They must not cache what they calculate. Instead, we have to introduce `StickyList` and `StickyMap`, which will do what we need to prevent multiple recalculations.

yegor256:
I'll fix it now

yegor256:
fixed in #186

PR #186, opened 2017-06-20
#180: StickyMap and StickyList

yegor256:
per #180:

  * `StickyMap` added
  * `StickyList` added
  * decorators are not sticky any more


yegor256:
@rultor merge

Issue #190, opened 2017-06-21
MapEntry


yegor256:
Let's create a simple class `MapEntry`, similar to `AbstractMap.SimpleEntry`.

fabriciofx:
@yegor256 Could you close this PR, please?

yegor256:
@fabriciofx thanks!

0crat:
Oops! Job `gh:yegor256/cactoos#190` was not in scope

PR #210, opened 2017-06-23
#190: MapEntry

fabriciofx:
As per #190 

`MapEntry` introduced.

### DIFF
--- a/src/main/java/org/cactoos/list/MapEntry.java
+++ b/src/main/java/org/cactoos/list/MapEntry.java
@@ -1,0 +1,100 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.list;
+
+import java.util.Map;
+
+/**
+ * MapEntry as {@link java.util.AbastractMap.Entry}.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @param <K> Key type
+ * @param <V> Value type
+ * @since 0.9
+ */
+public final class MapEntry<K, V> implements Map.Entry<K, V> {
+    /**
+     * The key.
+     */
+    private final K key;
+
+    /**
+     * The value.
+     */
+    private final V value;
+
+    /**
+     * Ctor.
+     * @param key The key of the entry's map.
+     * @param value The value associated to the key of the entry's map.
+     */
+    public MapEntry(final K key, final V value) {
+        this.key = key;
+        this.value = value;
+    }
+
+    @Override
+    public K getKey() {
+        return this.key;
+    }
+
+    @Override
+    public V getValue() {
+        return this.value;
+    }
+
+    @Override
+    // @checkstyle HiddenFieldCheck (1 line)
+    public V setValue(final V value) {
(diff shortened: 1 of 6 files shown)


0crat:

src/test/java/org/cactoos/list/MapEntryTest.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.9
+ * @checkstyle JavadocMethodCheck (500 lines)
+ */
+public final class MapEntryTest {
+
+    @Test
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
```
yegor256: @fabriciofx why do you need to use the same literals in all test methods?

yegor256: @fabriciofx use variables instead

src/test/java/org/cactoos/list/MapEntryTest.java
```diff
@@ -0,0 +1,85 @@
...
+    public void getKey() {
+        MatcherAssert.assertThat(
+            "Can't get key in the map entry",
+            new MapEntry<>("hello", "world").getKey(),
+            Matchers.equalTo("hello")
+        );
+    }
+
+    @Test
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
```
yegor256: @fabriciofx same here, use variables

src/test/java/org/cactoos/list/MapEntryTest.java
```diff
@@ -0,0 +1,85 @@
...
+        );
+    }
+
+    @Test(expected = UnsupportedOperationException.class)
+    public void cantSetValue() {
+        new MapEntry<>("one", "two").setValue("three");
+    }
+
+    @Test
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
```
yegor256: @fabriciofx same here

src/main/java/org/cactoos/list/MapEntry.java
```diff
@@ -0,0 +1,100 @@
...
+        return this.key;
+    }
+
+    @Override
+    public V getValue() {
+        return this.value;
+    }
+
+    @Override
+    // @checkstyle HiddenFieldCheck (1 line)
```
yegor256: @fabriciofx what this is all about?

yegor256:
@fabriciofx see above

src/main/java/org/cactoos/list/MapEntry.java
fabriciofx: @yegor256 I don't know why, qulice/checkstyle point an error because the method's parameter `value` has the same name of attribute `value`. If I'll change the code to:
```java
public V setValue(final V val)  {...}
```
the code works fine. So, I've two options:
1. To put the comment to qulice/checkstyle as above; or
2. To change the parameter's name.

I choosed option 1), because I'd like to get more semantic possible. WDYT?

yegor256: @fabriciofx just change the name. It's a bad practice to give method parameters same names as object attributes.

fabriciofx: @yegor256 I'll change it, but could you explain why is it a bad practice?

yegor256: @fabriciofx because it creates extra mess

yegor256: @fabriciofx it makes code less readable. How do I know which `value` do you mean in the code? `this.value` or just `value`?

src/main/java/org/cactoos/list/MapEntry.java
```diff
@@ -0,0 +1,100 @@
...
+     * The value.
+     */
+    private final V value;
+
+    /**
+     * Ctor.
+     * @param key The key of the entry's map.
+     * @param value The value associated to the key of the entry's map.
+     */
+    public MapEntry(final K key, final V value) {
```
yegor256: @fabriciofx the same issue here. It's a bad practice. This code is not very readable: `this.key = key`. Looks like "WTF?":)

yegor256:
@fabriciofx still some issues above

src/main/java/org/cactoos/list/MapEntry.java
fabriciofx: @yegor256 IMHO, it's very clear: `this.value` refers to an object's attribute. `value` refers to a parameter or a local variable. Besides, when seeing the interfaces/classes signatures is more clear to understand. For example, compare `MapEntry(final K key, final V value)` with `MapEntry(final K ky, final V vl)` what's easier to understand? Do you see my point?

src/main/java/org/cactoos/list/MapEntry.java
fabriciofx: @yegor256 See my comment above. But I'll change them, as requested. Thanks for your attention!

src/test/java/org/cactoos/list/MapEntryTest.java
fabriciofx: @yegor256 Ok.

src/test/java/org/cactoos/list/MapEntryTest.java
fabriciofx: @yegor256 Done.

src/test/java/org/cactoos/list/MapEntryTest.java
fabriciofx: @yegor256 Done.

src/main/java/org/cactoos/list/MapEntry.java
ixmanuel: @yegor256 Maybe in some of your projects one developer introduced a bug with this convention and thus, you decided to avoided it, but maybe in that time the objects were bigger than in these days that the objects are almost reduced to one method.  Also 

> Abbreviations, especially non-standard ones, are effectively terms-of-art, because understanding depends on correctly translating them into their non-abbreviated forms."
 
and thus, 

> The intended meaning for any abbreviation you use should be easily found by a web search.

[https://swift.org/documentation/api-design-guidelines/](url)

yegor256:
@rultor merge

Issue #216, opened 2017-06-23
Range


VsSekorin:
Let's create `IntRange`. Range of integer number (Z).

0crat:
@vssekorin I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

PR #217, opened 2017-06-23
#216 Range

VsSekorin:
As per #216 
`IntRange` inroduced

### DIFF
--- a/src/main/java/org/cactoos/list/IntRange.java
+++ b/src/main/java/org/cactoos/list/IntRange.java
@@ -1,0 +1,64 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


0crat:

src/main/java/org/cactoos/list/IntRange.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Range of integer number (Z).
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Vseslav Sekorin (vssekorin@gmail.com)
+ * @version $Id$
+ * @since 0.8
+ */
+public final class IntRange implements Iterable<Long> {
```
fabriciofx: @VsSekorin Is it not strange a class called `IntRange` and it returns a `Long`?

VsSekorin: [Integer number (Z)](https://en.wikipedia.org/wiki/Integer). Long contains integer numbers.

fabriciofx: @VsSekorin Yeah, I know, but if I'm using `IntRange` I'd like get a range of `Integer`s, right? In other words, the problem is the name. Maybe `NumberRange` makes more senses. WDYT?

yegor256:
@VsSekorin it seams that we can use `new LimitedIterable(new SkippedIterable(new NaturalNumbers()))` to achieve the same goal, right?

VsSekorin:
@yegor256 Yeah

Issue #56, opened 2017-05-30
Text is not Comparable<Text>


yegor256:
Let's make sure that `Text extends Comparable<Text>`

fabriciofx:
We should use lombok to implement `compareTo()`?

englishman:
@fabriciofx it'll introduce a heavy external dependency, I guess. In README.md is:

> How to use. **The library has no dependencies.** All you need is this (get the latest version here):

fabriciofx:
@englishman Yeah, you're correct. Thank you for the clarification.

fabriciofx:
@englishman I'm trying solve this issue. At the moment, my `compareTo()` method is:

```java
    public int compareTo(final Text text) {
        try {
            return this.asString().compareTo(text.asString());
        } catch (final IOException ex) {
            throw new IllegalArgumentException(ex);
        }
    }
```
So my question is: if something goes wrong, we should throws an `IllegalArgumentException` or `RuntimeException` encapsulating the `IOException`? WDYT?

yegor256:
@fabriciofx I think we should introduce `UncheckedScalar`, similar to `UncheckedFunc`: #74

englishman:
@fabriciofx don't clearly understand what can be wrong in this `compareTo` code. Do we need here any exception processing?

fabriciofx:
@yegor256 Could you check the latest comments in the PR?

fabriciofx:
Closed #219

yegor256:
@fabriciofx thanks!

PR #219, opened 2017-06-24
#56: Text extends Comparable<Text>

fabriciofx:
As per #56 let´s do `Text` extends `Comparable<Text>`.

### DIFF
--- a/src/main/java/org/cactoos/text/BytesAsText.java
+++ b/src/main/java/org/cactoos/text/BytesAsText.java
@@ -114,4 +114,9 @@ public final class BytesAsText implements Text {
         return new String(this.bytes.asBytes(), this.charset);
     }
 
+    @Override
+    public int compareTo(final Text text) {
+        return new UncheckedText(text).compareTo(this);
+    }
+
 }
--- a/src/main/java/org/cactoos/text/BytesAsText.java
+++ b/src/main/java/org/cactoos/text/BytesAsText.java
@@ -116,7 +116,7 @@ public final class BytesAsText implements Text {
 
     @Override
     public int compareTo(final Text text) {
-        return new UncheckedText(text).compareTo(this);
+        return new UncheckedText(this).compareTo(text);
     }
 
 }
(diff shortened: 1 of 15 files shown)


0crat:

src/main/java/org/cactoos/text/BytesAsText.java
```diff
@@ -114,4 +114,9 @@ public String asString() throws IOException {
         return new String(this.bytes.asBytes(), this.charset);
     }
 
+    @Override
+    public int compareTo(final Text text) {
+        return new UncheckedText(text).compareTo(this);
```
alex-semenyuk: @fabriciofx A little bit confused order - we compare argument with this , but in original variant is vice versa.  WDYT?

fabriciofx: @alex-semenyuk We need wrap `text` into `UncheckedText` to not throws an exception and after that, call `Unchecked`s `compareTo` to compare itself. How all `Text` implementations are immutable, I don't see any problem.  The reason it's because `UncheckedText` has the real comparing between strings and to avoid to replicate this code I did this way. But I'm opened to suggestions! :)

alex-semenyuk: @fabriciofx I understand that it will lead to replicate the code, but when I read 

> returns a negative integer, zero, or a positive integer as this object is less than, equal to, or greater than the specified object.

 I wait that the result will be same as in this contract or at least the same for all classes include `UncheckedText`

fabriciofx: @alex-semenyuk I agree with you, but the contract wasn't broken. The difference is who will do the compare is `UncheckedText`. Anyway, the alternative do something like:
```java
return new UncheckedText(this).asString().compareTo(new UncheckedText(text).asString());
```
Please, not I can't call `this.asString()` or `text.asString()` inside `compareTo()` because both throw exceptions. So, I need wrap them with `UncheckedText` before call `asString()`. Makes sense?

alex-semenyuk: @fabriciofx Maybe something like this `new UncheckedText(this).compareTo(text)` better?

fabriciofx: @alex-semenyuk Makes sense and you approach is better. I'l change them. Thanks!

yegor256:
@rultor merge

Issue #221, opened 2017-06-24
Abs


fabriciofx:
Let's implements `Abs`, that returns the same of `Math.abs()`.

0crat:
@fabriciofx I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

memoyil:
ok i'll do it

fabriciofx:
@memoyil I'm sorry, but I've finished it. Bu I'll appreciate your review! :)

fabriciofx:
@yegor256 Could you assign this issue to me and review my PR? Thanks in advance!

fabriciofx:
@yegor256 ping!

yegor256:
@fabriciofx I don't see a big value of replacing `Math.abs()` with a class. It's a very very primitive procedure, no need to re-write it.

0crat:
Job `gh:yegor256/cactoos#221` is not assigned, can't get start time

0crat:
Job [#221](https://github.com/yegor256/cactoos/issues/221) was not in scope

PR #222, opened 2017-06-24
#221: IntAbs and RealAbs

fabriciofx:
As per #221 :

`IntAbs` introduced.
`RealAbs` introduced.

### DIFF
--- a/src/main/java/org/cactoos/math/IntAbs.java
+++ b/src/main/java/org/cactoos/math/IntAbs.java
@@ -1,0 +1,98 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.math;
+
+import org.cactoos.Scalar;
+import org.cactoos.func.UncheckedScalar;
+
+/**
+ * Abs of a integer number.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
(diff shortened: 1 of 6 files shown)


0crat:

memoyil:
Incase java Math library is changed, we should not depend Math library i think. 
Maybe something like that
```
if (number < 0) {
    number *= -1;
}
```

or 

`abs_number = (number < 0) ? -number : number;
`

src/main/java/org/cactoos/math/IntAbs.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Abs of a integer number.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.9
+ */
+public final class IntAbs implements Scalar<Long> {
```
VsSekorin: You yourself wrote me about `Int* implements Scalar<Long>`. #217

src/main/java/org/cactoos/math/IntAbs.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Scalar.
+     */
+    private final UncheckedScalar<Long> scalar;
```
VsSekorin: Why not just `Scalar<Long>`?

src/main/java/org/cactoos/math/IntAbs.java
fabriciofx: @VsSekorin Yeah, and I was wrong, because I'm thinking in the `Integer` type and not in terms of [Integer set](https://en.wikipedia.org/wiki/Integer). So, makes sense an `IntAbs` or a `IntRange`.

src/main/java/org/cactoos/math/IntAbs.java
fabriciofx: @VsSekorin Just to avoid create a `UncheckedScalar` inside the `asValue` method. It performs similar to [FileAsInput](https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/io/FileAsInput.java)

src/main/java/org/cactoos/math/RealAbs.java
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Abs of a real number.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.9
+ */
+public final class RealAbs implements Scalar<Double> {
```
alex-semenyuk: @fabriciofx Maybe make sense to do it parameterized and not need for this two classes, need just one? WDYT?

fabriciofx: @alex-semenyuk Yeah, it's a good approach, but the problem is `Math.abs(this.scalar.asValue())`. If I changed to something like: `public final class Abs<T extends Number> implements Scalar<T>`, this.scalar.asValue()` won't work for a generic parameter `T`. Unless you've a workaround for this problem... :)

fabriciofx:
@memoyil I think the probability of `Math.abs()` change is very tiny, to don't say zero... :) But thanks for your review!

yegor256:
@fabriciofx do we really need to move all `Math.*` methods to objects? It seems like an over-kill to me. Very few (or nobody) will ever use our objects. Don't you think?

memoyil:
@yegor256 Yegor i think we should make an histogram or smth. that shows the most frequent usage of math and string methods. Then we should only implement this methods in cactoos. No more no less

fabriciofx:
@yegor256 I really don't know. I thought was a good idea implement these methods like objects. But the @memoyil idea seems very interesting...

yegor256:
@memoyil yes, sounds like a plan. We definitely don't want to turn all static methods from JDK into our objects. Maybe let's create a summary spreadsheet where all, say, Guava methods are listed and our alternatives are in front of them.

yegor256:
@fabriciofx @memoyil check this out: c5defd3

llorllale:
@fabriciofx #221 was closed, see [this](https://github.com/yegor256/cactoos/issues/221#issuecomment-361256571).

Please close this PR

0crat:

PR #263, opened 2017-06-28
172: Max and Min

fabriciofx:
As per #172:

- `Max` introduced.
- `Min` introduced.


svendiedrichsen:
this Min/Max cannot handle empty lists/iterators

fabriciofx:
@svendiedrichsen Thanks for your report. I think fixed them now. Could you check it?

yegor256:
@rultor merge

svendiedrichsen:
Tests are looking good. I was just thinking if the original NoSuchElementException isn't already good enough. WDYT?

fabriciofx:
@svendiedrichsen Makes sense. I'll do it. Thanks!

Issue #269, opened 2017-06-29
Change exception type in Max and Min to a proper exception


fabriciofx:
Let's change `IOException` in `Max` and `Min` to `NoSuchElementException`.

0crat:
@yegor256 please, pay attention to this issue

fabriciofx:
Closed #270 .

PR #270, opened 2017-06-29
#269: Change Max and Min to a proper exception

fabriciofx:
As per #269 .

### DIFF
--- a/src/main/java/org/cactoos/list/Max.java
+++ b/src/main/java/org/cactoos/list/Max.java
@@ -66,7 +66,7 @@ public final class Max<T extends Comparable<T>> implements Scalar<T> {
         final Iterator<Scalar<T>> iter = this.items.iterator();
         if (!iter.hasNext()) {
             throw new NoSuchElementException(
-                "Can't find greater element in an empty interable"
+                "Can't find greater element in an empty iterable"
             );
         }
         T max = iter.next().value();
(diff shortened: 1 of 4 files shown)


0crat:
@yegor256 please, pay attention to this pull request

svendiedrichsen:
@fabriciofx: Sorry to annoy you again but if you call next() on an empty iterator the NoSuchElementException will be thrown automatically. You don't need to check for hasNext() and throw it yourself. Thus you can remove the if statement.

fabriciofx:
@svendiedrichsen Don't apologize. I'm who should be thanking you, because it's one more opportunity to learn.  The problem is `next()` throws `NoSuchElementException` without a proper message. So, I think better in this case throws a `NoSuchElementException` with a message like "Can't find greater element in an empty interable". WDYT?

svendiedrichsen:
I agree. Having a proper message is better.

src/main/java/org/cactoos/list/Max.java
```diff
@@ -65,7 +65,9 @@ public Max(final Iterable<Scalar<T>> items) {
     public T value() throws Exception {
         final Iterator<Scalar<T>> iter = this.items.iterator();
         if (!iter.hasNext()) {
-            throw new IOException("Iterable is empty");
+            throw new NoSuchElementException(
+                "Can't find greater element in an empty interable"
```
alex-semenyuk: @fabriciofx Typo "interable"

fabriciofx: @alex-semenyuk I'll fix it! Thanks!

yegor256:
@rultor merge

Issue #287, opened 2017-07-04
Range for NaturalNumbers


fabriciofx:
Let's add range support to `NaturalNumbers`. I'll handle it.

fabriciofx:
Closed #288.

PR #288, opened 2017-07-04
#287 Range for NaturalNumbers

fabriciofx:
As per #287.

### DIFF
--- a/src/main/java/org/cactoos/list/NaturalNumbers.java
+++ b/src/main/java/org/cactoos/list/NaturalNumbers.java
@@ -25,6 +25,8 @@ package org.cactoos.list;
 
 import java.util.Iterator;
 import java.util.stream.LongStream;
+import org.cactoos.Scalar;
+import org.cactoos.func.UncheckedScalar;
 
 /**
  * Iterable providing Natural Numbers.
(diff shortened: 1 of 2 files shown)


src/test/java/org/cactoos/list/NaturalNumbersTest.java
```diff
@@ -56,4 +56,25 @@ public void notStartsWithNegativeNumbers() {
             new ScalarHasValue<>(0L)
         );
     }
+
+    @Test
+    public void containsNaturalNumbersRange() {
+        MatcherAssert.assertThat(
+            "Can't get range of natural numbers",
+            // @checkstyle MagicNumber (2 lines)
```
VsSekorin: This string is repeated three times in this class. Perhaps this should be moved to class javadoc? Like

> checkstyle JavadocMethodCheck (500 lines)

src/main/java/org/cactoos/list/NaturalNumbers.java
```diff
@@ -35,8 +35,43 @@
  */
 public final class NaturalNumbers implements Iterable<Long> {
 
+    /**
+     * First natural number.
+     */
+    private final long first;
```
fabriciofx: @VsSekorin Ok. I'll change it. Thanks!

src/test/java/org/cactoos/list/NaturalNumbersTest.java
fabriciofx: @VsSekorin I don't think so. Will it aggregate value to this software? WDYT?

yegor256:
@rultor merge

Issue #289, opened 2017-07-04
Length of text


VsSekorin:
Let's create `LengthOfText`

yegor256:
@VsSekorin what's wrong with `.toString().length()`?

andriyshevchenko:
It's too procedural.

VsSekorin:
@yegor256 I think this is more object-oriented. I create an object Scalar that is the length of the text.
```
Scalar<Integer> length = new LengthOfText(
        new TextDcr(arg)
    );
```
or
```
Scalar<Integer> length = new LengthOfText(text);
```
Instead of taking the string from the text, then take its length, and then wrap the result in a scalar.
```
Scalar<Integer> length = new Scalar<>() {
        @Override
        public Integer value() throws Exception {
            return new TextDcr(
                arg
            ).asString().length();
        }
    };
```
or
```
Scalar<Integer> length = new Scalar<>() {
        @Override
        public Integer value() throws Exception {
            return text.asString().length();
        }
    };
```

Also I always use only primitive cactoos, and do not jump Scalar -> Integer -> Scalar.

0crat:
Oops! Job `gh:yegor256/cactoos#289` was not in scope

PR #290, opened 2017-07-04
#289 Length of text

VsSekorin:
As per #289 
`LengthOfText` introduced

### DIFF
--- a/src/main/java/org/cactoos/text/LengthOfText.java
+++ b/src/main/java/org/cactoos/text/LengthOfText.java
@@ -1,0 +1,66 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


src/test/java/org/cactoos/text/LengthOfTextTest.java
```diff
@@ -0,0 +1,60 @@
...
+            ).value(),
+            Matchers.equalTo(4)
+        );
+    }
+
+    @Test
+    public void valueStringAsArg() throws Exception {
+        MatcherAssert.assertThat(
+            new LengthOfText(
+                new StringAsText("string")
```
fabriciofx: @VsSekorin Excuse me, but it shouldn't be just `"string"` here?

VsSekorin: My bad. Thanks.

src/test/java/org/cactoos/text/LengthOfTextTest.java
```diff
@@ -51,10 +51,26 @@ public void valueTextAsArg() throws Exception {
     @Test
     public void valueStringAsArg() throws Exception {
         MatcherAssert.assertThat(
+            new LengthOfText("string").value(),
+            Matchers.equalTo(6)
```
fabriciofx: @VsSekorin I'm sorry, but what do you think use `ScalarHasValue` instead of `Matchers.equalTo()`?

src/test/java/org/cactoos/text/LengthOfTextTest.java
```diff
@@ -0,0 +1,76 @@
...
+ * SOFTWARE.
+ */
+package org.cactoos.text;
+
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+/**
+ * Test case for {@link org.cactoos.Text}.
```
alex-semenyuk: Typo in link, should be LengthOfText

src/main/java/org/cactoos/text/LengthOfText.java
```diff
@@ -0,0 +1,66 @@
...
+     * Ctor.
+     * @param text The text
+     */
+    public LengthOfText(final Text text) {
+        this.origin = text;
+    }
+
+    @Override
+    public Integer value() throws Exception {
+        return this.origin.asString().length();
```
alex-semenyuk: maybe better 

> new UncheckedText(this.origin)

 then can be removed throws Exception

src/test/java/org/cactoos/text/LengthOfTextTest.java
VsSekorin: Thank you. I forgot about them.

src/main/java/org/cactoos/text/LengthOfText.java
VsSekorin: Explain a little more in detail what is the problem with the exception in this class, please.

yegor256:
@rultor merge

Issue #143, opened 2017-06-15
CallableWithFallback


yegor256:
Let's create `CallableWithFallback`, similar to `RunnableWithFallback` we introduced in #141

cheparinV:
@yegor256 i'll handle this issue.

ashtonhogan:
@cheparinV seems this was merged, can you please close this issue?

yegor256:
@cheparinV @ashtonhogan thanks

0crat:
The job is not in WBS, won't close the order

PR #291, opened 2017-07-05
#143 CallableWithFallback

cheparinV:
As per #143

### DIFF
--- a/src/main/java/org/cactoos/func/CallableWithFallback.java
+++ b/src/main/java/org/cactoos/func/CallableWithFallback.java
@@ -1,0 +1,71 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


src/main/java/org/cactoos/func/CallableWithFallback.java
```diff
@@ -0,0 +1,71 @@
...
+ * SOFTWARE.
+ */
+package org.cactoos.func;
+
+import java.util.concurrent.Callable;
+import org.cactoos.Proc;
+
+/**
+ * Runnable with a fallback plan.
+ *
```
alex-semenyuk: @cheparinV Typo "Callable"

src/main/java/org/cactoos/func/CallableWithFallback.java
```diff
@@ -0,0 +1,71 @@
...
+
+    /**
+     * The callable.
+     */
+    private final Callable<Y> callable;
+
+    /**
+     * The fallback.
+     */
+    private final Proc<Throwable> fallback;
```
yegor256: @cheparinV why not `Func<Throwable, Y>`?

yegor256:
@cheparinV I would use `Func` instead, see my comment above

yegor256:
@rultor merge

Issue #294, opened 2017-07-07
NoNulls for Bytes


fabriciofx:
Let's implement `NoNulls` for `Bytes`. I'll handle it.

fabriciofx:
Done in #295.

PR #295, opened 2017-07-07
#294: NoNulls for Bytes

fabriciofx:
As per #294 .

- `Bytes.NoNulls` introduced.


yegor256:
@fabriciofx why `IOException`? I would rather use `IllegalArgumentException` or `IllegalStateException`, depending on the situation.

fabriciofx:
@yegor256 Right. I'll fix them. Thanks!

yegor256:
@rultor mege

Issue #274, opened 2017-07-02
Brevity: Renaming and Refactoring


ixmanuel:
Is it pertinent to question the naming of these objects?
- ArrayAsIterable
- IterableAsList   
- IterableAsMap 

_In the process of writing these objects with these names is a pleasure, but, to read them? I think we could opt for brevity._

Do these objects/concepts lose their meaning? if replaced by:
- ArrayOf
- ListOf
- MapOf

Let's see them in context to see if they reduce their clarity:
```java
new MappedIterable<>(
    new ArrayOf<>("how", "are", "you")
);

new CycledIterable<>(
    new ArrayOf<>()
);

new ListOf<>(-1, num, 0, 1);

new MapOf<>(
    new MappedIterable<Map.Entry<?, ?>, Map.Entry<String, String>>(
        new ArrayOf<>(entries),
        input -> new MapEntry<>(...)
    )
);
```

**Can an array or list become non-iterable?**

If the previous arguments have some kind of truthiness, then, could we shorten the rest?

```java
new Mapped<>(
    new ArrayOf<>("how", "are", "you")
);

new Cycled<>(
    new ArrayOf<>()
);

new ListOf<>(-1, num, 0, 1);

new MapOf<>(
    new Mapped<Map.Entry<?, ?>, Map.Entry<String, String>>(
        new ArrayOf<>(entries),
        input -> new MapEntry<>(...)
    )
);
```

0crat:
@yegor256 please, pay attention to this issue

ixmanuel:
@yegor256, your idea of multiple constructors can simplify this API into a highly polymorphic solution. Thus, it could be reduced the quantity of object-names for the end-developer-user. 

```
StringAsText    |    TextOf
BytesAsText     |    TextOf

FileAsInput     |    InputOf
ResourceAsInput	|    InputOf
BytesAsInput    |    InputOf

InputAsBytes	|    BytesOf
TextAsBytes	|    BytesOf

LengthOfInput	|    LengthOf
LengthOfIterable|    LengthOf

IterableAsList	|    ListOf
ArrayAsIterable	|    ArrayOf
FilteredIterable|    Filtered
MappedIterable	|    Mapping

ProcAsFunc	|    FuncOf
```

Also asBytes() can be renamed just for bytes()
```java
byte[] data = new BytesOf(
  new InputOf("foo/img.jpg")
).bytes();
```

[...]

PR #312, opened 2017-07-12
Brevity#274 input of

ixmanuel:
Trial for #274 - InputOf


src/main/java/org/cactoos/io/InputOf.java
```diff
@@ -0,0 +1,201 @@
...
+     */
+    private final Input origin;
+
+    /**
+     * Ctor.
+     *
+     * @param src The file
+     */
+    public InputOf(final File src) {
+        this(new FileAsInput(src));
```
fabriciofx: @ixmanuel You can't use lambda function here? I thought in something like:
```
public InputOf(final File src) {
  this(src -> new FileInputStream(src));
}
```

src/main/java/org/cactoos/text/RegexText.java
```diff
@@ -0,0 +1,78 @@
+/**
```
fabriciofx: @ixmanuel WDYT to create a new PR only with Regex classes code?

src/test/java/org/cactoos/text/RegexTextTest.java
```diff
@@ -0,0 +1,58 @@
+/**
```
fabriciofx: @ixmanuel The same above.

src/main/java/org/cactoos/text/RegexText.java
ixmanuel: Thanks. I've already fixed it.

src/test/java/org/cactoos/text/RegexTextTest.java
ixmanuel: Thanks. I've already fixed it.

src/main/java/org/cactoos/io/InputOf.java
ixmanuel: Yes, I tried according to the example that you exposed in the conversation # 312 but I didn't see more benefits for the rest of the cases, so I went back to the idea of Decorating Envelopes and thus, I didn't change anything, I have only created InputOf.

fabriciofx:
@ixmanuel Could you resubmit `RegexText` as a new issue/PR? Thanks in advance!

yegor256:
@ixmanuel I don't like that `new InputOf(String)` would work as a `ResourceAsInput`. That's counter intuitive. I would expect it to do what `BytesAsInput` is doing now: convert a string content to the `Input`.

yegor256:
@ixmanuel this is what I would expect:

  * `InputOf(File)`
  * `InputOf(Path)`
  * `InputOf(URL)`
  * `InputOf(URI)`
  * `InputOf(String)`
  * `InputOf(Text)`
  * `InputOf(Bytes)`
  * `InputOf(byte[])`
  * `InputOf(char[])`
  * `InputOf(StringBuilder)`
  * `InputOf(StringBuffer)`
  
maybe something else. "Resource" is not an option. We still need to have `ResourceAsInput`.

ixmanuel:
Yes, removing ResourceAsInput is counter intuitive.

ixmanuel:
@yegor256 InputOf done.

src/main/java/org/cactoos/io/InputOf.java
```diff
@@ -0,0 +1,258 @@
...
+    public InputOf(final InputStream stream) {
+        this(() -> stream);
+    }
+
+    /**
+     * Ctor.
+     *
+     * @param input The input
+     */
+    public InputOf(final Input input) {
```
yegor256: @ixmanuel let's make it `private`

yegor256:
@ixmanuel looks good, one small comment above

ixmanuel:
@yegor256, Now the main constructor is private.

yegor256:
@rultor try to merge

ixmanuel:
@yegor256, new test and fixed an independent ctor.

yegor256:
@ixmanuel but the branch is merged already! Or I don't understand what's going on :)

ixmanuel:
@yegor256 I updated this almost at the same time you published it. Then I added some more improvements. Let me close this and open another one. Thanks.

ixmanuel:
@yegor256 Note: If you are reviewing now, BytesOf #340 and ReplacedAllText #341 are in synch with the branch you merged not with these post-updates. Now I'm creating a new pull request for theses post-updates of InputOf.

0crat:
Oops! Job `gh:yegor256/cactoos#312` was not in scope

Issue #314, opened 2017-07-14
RegexText


ixmanuel:
Returning a valid regular expression.

Here it is the request, thanks @fabriciofx.
> @ixmanuel Could you resubmit RegexText as a new issue/PR? Thanks in advance!

voddan:
Is this issue being worked on? I would like to take it if it is not

ixmanuel:
@voddan yes, it is on, you can take it. You only have to remove the RegexText and everything will work without any issues. Am I right @yegor256 ?

ixmanuel:
@voddan, well, you'll have to change some things like ArrayAsIterable to IterableOf because it is outdated with the recent refactoring.

ixmanuel:
@voddan Sorry, I read it wrong. What you want is to revive RegexText. I don't know, it will be removed from ReplacedAllText but it doesn't mean that RegexText::match is wrong so I think you can reopen it.

voddan:
@ixmanuel Sorry, I am new, could you please explain what you mean by "reviving RegexText"? Right now there is no such class in the master branch. Did it existed at some point? Could you post the number of the commit which you think I should take a look at?

ixmanuel:
@voddan you can see how it was in https://github.com/yegor256/cactoos/pull/315. I closed it because I was not sure if this kind of thinking (what you can see in the previous link) has space in this project. The discussion wasn't ended it but you can see there the concerns.

voddan:
Unfortunately I can't continue working on this ticket. My vacation went sideways ;)

0crat:
There is an unrecoverablefailure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.ClassCastException: org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(JsonObjectBuilderImpl.java:252)
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
	at com.zerocracy.radars.github.RbByLabel.react(RbByLabel.java:82)
	at com.zerocracy.radars.github.RbByActions.react(RbByActions.java:65)
	at com.zerocracy.radars.github.Rebound$Chain.react(Rebound.java:74)
[...]

PR #315, opened 2017-07-14
Regex text#314

ixmanuel:
#314


src/main/java/org/cactoos/text/RegexText.java
```diff
@@ -0,0 +1,79 @@
...
+import org.cactoos.Text;
+
+/**
+ * Returns a valid regex expresion.
+ *
+ * @author Ix (ixmanuel@yahoo.com)
+ * @version $Id$
+ * @since 0.12
+ */
+public final class RegexText implements Text {
```
fabriciofx: @ixmanuel What do you think add a method to check if a string matches with this regex?

ixmanuel: Could we review it together? because I think that if we have a compareTo() why not a matches(). Although, the underlying concept in RegexText is to create a Type. It has the same meaning as StringAsText (TextOf sounds more like a type) because in Java there is not a real type for Regex, it relies on a string. 

Some refections:

1. RegextText was conceived to distinguish a simple text from a regex and also to verify its syntax. The difference with StringAsText is that the string is statically verified while RegexText not. In the proposal ReplacedAllText it uses ReplacedRegexText that also uses RegexText.
2. If we implement matches() in RegextText we could be tempted to add more methods like: Split, ReplacedFirst and ReplacedAll but, matches is reasonable, because we could say that it is its natural behaviour.
3. Once added matches() we have to create a new protocol and thus, RegexText is not more a  consistent decoration of Text.

These previous ideas are what they are, just ideas. I can't visualise a big issue and I think we could add it. So let me present you an implementation in order to be sure that we are understanding the same.

```java
    @Override
    public boolean matchesIn(final Text text) {
        return this.origin.matches(text.asString());
    }
```
with a protocol: Match?

ixmanuel: Assessment:

```java
new RegexText("[abc]").matched(new Text("abc"));
new RegexText("[abc]").matcher(new Text("abcabcabc"));
new RegexText("[abc]").matches(new Text("abcabcabc"));
```
--- vs  ---
```
new Matched(
    new RegexText("[abc]"), new Text("abc")
)->value();

new Matches(
    new RegexText("[abc]"), new Text("abracadabra")
)->iterator();
```

ixmanuel:
@fabriciofx ReplacedAllText was merged. I'm closing this to avoid duplication, but I'll keep the dialogue in the issue.

0crat:
Oops! Job `gh:yegor256/cactoos#315` was not in scope

Issue #225, opened 2017-06-24
Porting Takes codecs to Cactoos


fabriciofx:
Let's port [Takes](http://www.takes.org) [codecs](https://github.com/yegor256/takes/tree/master/src/main/java/org/takes/facets/auth/codecs) to Cactoos. @yegor256 WDYT?

0crat:
@fabriciofx I'm not managing this repo, remove the [webhook](https://github.com/yegor256/cactoos/settings/hooks) or contact me in [Slack](http://www.zerocracy.com) //cc @yegor256

memoyil:
Great Idea !

fabriciofx:
@memoyil This idea was inspired by your work in `Base64DecodedText`. Do you like to do this port? If yes, I suggest you to create a pair of issue/PR to each codec. WDYT?

memoyil:
@fabriciofx  I am OK but I think we should wait @yegor256 ideas about this issue.

fabriciofx:
@memoyil Yeah, I agree. Let's wait.

yegor256:
@fabriciofx yes, definitely, let's move them in here.

fabriciofx:
@yegor256 Ok. @memoyil Would you like implement this porting? :)

memoyil:
@fabriciofx OK, I will move them here

fabriciofx:
@memoyil Any news about this issue?

memoyil:
@fabriciofx I will have a commit today.

fabriciofx:
@memoyil Ok. Thanks!

memoyil:
@fabriciofx Hello I created a pull request. I put 2 implementations. When we made a compromise on these I can implement all of them.

memoyil:
@fabriciofx @yegor256 We should close this issue.

fabriciofx:
@memoyil @yegor256 WDYT close this issue? And who will earn the points?

llorllale:
@fabriciofx Has this been done? I could not quite follow which (or all?) of the codecs you wanted ported. Can this issue be closed?

llorllale:
@fabriciofx can we close this?

llorllale:
@fabriciofx ping

llorllale:
@fabriciofx closed due to inactivity. Feel free to request this feature in another ticket.

0crat:
The job is not in WBS, won't close the order

PR #320, opened 2017-07-18
#225: Porting takes into cactoos

memoyil:
Porting takes into cactoos


src/main/java/org/cactoos/Codec.java
```diff
@@ -0,0 +1,63 @@
...
+import java.io.IOException;
+
+/**
+ * Codec.
+ *
+ * <p>All implementations of this interface must be immutable and thread-safe.
+ *
+ * @author Mehmet Yildirim (memoyil@gmail.com)
+ * @version $Id$
+ * @since 0.1
```
fabriciofx: @memoyil We already are in the 0.11.1 version.

src/main/java/org/cactoos/Codec.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public interface Codec {
```
fabriciofx: @memoyil I was thinking in something more generic, like `Code<X, Y>`, WDYT?

src/main/java/org/cactoos/Codec.java
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Encode text into bytes.
+     * @param text The text
+     * @return Text
+     * @throws IOException If fails
+     */
+    byte[] encode(String text) throws IOException;
```
fabriciofx: @memoyil So, it will be something like `Y encode(X input) throws IOException;`

src/main/java/org/cactoos/Codec.java
```diff
@@ -0,0 +1,63 @@
...
+     * {@link org.cactoos.codec.DecodingException}, if it's not
+     * possible to decode the incoming byte array. This exception will mean
+     * that the user can't be authenticated and {@code Identity.ANONYMOUS}
+     * object will be identified.
+     *
+     * @param bytes Text
+     * @return Identity
+     * @throws IOException If fails
+     */
+    Text decode(byte[] bytes) throws IOException;
```
fabriciofx: @memoyil An here, the opposite: `X encode(Y input) throws IOException;`

src/main/java/org/cactoos/codec/Base64Codec.java
```diff
@@ -0,0 +1,98 @@
...
+    public byte[] encode(final String text) throws IOException {
+        return Base64.getEncoder().encode(this.origin.encode(text));
+    }
+
+    @Override
+    public Text decode(final byte[] bytes) throws IOException {
+        final byte[] illegal = Base64Codec.checkIllegalCharacters(bytes);
+        if (illegal.length > 0) {
+            throw new DecodingException(
+                String.format(
```
fabriciofx: @memoyil Please, use `FormattedText` instead.

src/main/java/org/cactoos/codec/GzipCodec.java
```diff
@@ -0,0 +1,90 @@
...
+    /**
+     * Ctor.
+     * @param codec Original
+     */
+    public GzipCodec(final Codec codec) {
+        this.origin = codec;
+    }
+
+    @Override
+    public byte[] encode(final String text) throws IOException {
```
fabriciofx: @memoyil I was thinking if `GzipCodec` could works with *anything* not just `String`. WDYT in something like `Bytes encode(Bytes input) throws IOException`?

src/main/java/org/cactoos/codec/GzipCodec.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            gzip.write(this.origin.encode(text));
+        } finally {
+            gzip.close();
+        }
+        return out.toByteArray();
+    }
+
+    @Override
+    public Text decode(final byte[] bytes) throws IOException {
```
fabriciofx: @memoyil See above.

src/main/java/org/cactoos/Codec.java
memoyil: @fabriciofx I have some questions over there. If we do like that how could we possible to detect decorated codec type? And pass the correct type to their encode/decode?

src/main/java/org/cactoos/codec/Base64Codec.java
```diff
@@ -0,0 +1,98 @@
...
+                    "Illegal character in Base64 encoded data. %s",
+                    Arrays.toString(illegal)
```
alex-semenyuk: @memoyil you can write something like 'new UncheckedText(new BytesAsText(illegal)).asString()' instead this

memoyil: @alex-semenyuk I did it by FormattedText

alex-semenyuk: I meant this one  `Arrays.toString(illegal)`.

memoyil: @alex-semenyuk Ok i fixed that olso

src/main/java/org/cactoos/Codec.java
```diff
@@ -0,0 +1,61 @@
...
+ */
+public interface Codec {
+
+    /**
+     * Encode text into bytes.
+     * @param input The input
+     * @return Bytes
+     * @throws IOException If fails
+     */
+    Bytes encode(Text input) throws IOException;
```
fabriciofx: @memoyil Sorry, but how can I encode or decode a `File`, for example? So, I as thinking in something like:
```java
public interface Code {
  Output encode(Input input);
  Output decode(Input input);
}
```
And to compact `Text` (for instance), you can do: ` new GzipCodec().encode(new BytesAsInput("my text"));` or something like that. I think you got the idea... :)

svendiedrichsen:
You might want to use try-with-resources for working with streams. Makes the code more compact and thus more readable.

yegor256:
@memoyil I don't really like the design here. It's purely procedural. The very name `Codec` means that it's a collection of encoding and decoding procedures. How about `InputAsBase64` and `Base64AsInput`?

memoyil:
@yegor256 Ok, I will refactor the code based on that.

yegor256:
@memoyil thanks!

svendiedrichsen:
Maybe it makes sense to have i.e. InputToBase64/Base64ToInput implement the Func interface.

memoyil:
@yegor256 I implemented the new approach. If you approve this I will continue the add codecs that way.

src/main/java/org/cactoos/codec/Base64IllegalAsBytes.java
```diff
@@ -0,0 +1,71 @@
...
+ * @author Mehmet Yildirim (memoyil@gmail.com)
+ * @version $Id$
+ * @since 0.12
+ */
+public final class Base64IllegalAsBytes implements Bytes {
+
+    /**
+     * All legal Base64 chars.
+     */
+    private static final String BASE64CHARS =
```
yegor256: @memoyil I didn't understand what is the point of this class. Just put this functionality into `Base64ToInput`

yegor256:
@memoyil looks good, just one comment above

memoyil:
@yegor256 Fixed it!

src/main/java/org/cactoos/codec/Base64ToInput.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public InputStream stream() throws IOException {
+        final InputStream stream = this.origin.stream();
+        final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        int read = stream.read();
+        while (read != -1) {
+            out.write(read);
+            read = stream.read();
+        }
+        final byte[] bytes = out.toByteArray();
```
yegor256: @memoyil you seem to be reimplementing what we already have in `InputAsBytes`. No need to read it here, just use `new InputAsBytes(this.origin).bytes()`

src/main/java/org/cactoos/codec/Base64ToInput.java
```diff
@@ -0,0 +1,106 @@
...
+     * Get the byte array for non-Base64 characters.
+     * @param bytes Byte Array
+     * @return An array of the found non-Base64 characters.
+     * @throws IOException If fails
+     */
+    private static byte[] getIllegalCharacters(final byte[] bytes) throws
+        IOException {
+        final ByteArrayOutputStream output = new ByteArrayOutputStream();
+        for (int pos = 0; pos < bytes.length; ++pos) {
+            if (BASE64CHARS.indexOf(bytes[pos]) < 0) {
```
yegor256: @memoyil I suspect that it has to be `> 0` instead. Are you sure this code works?

src/main/java/org/cactoos/codec/InputToBase64.java
```diff
@@ -0,0 +1,70 @@
...
+
+    @Override
+    public InputStream stream() throws IOException {
+        final InputStream stream = this.origin.stream();
+        final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        int read = stream.read();
+        while (read != -1) {
+            out.write(read);
+            read = stream.read();
+        }
```
yegor256: @memoyil again, don't re-invent this functionality, we have it already in `InputAsBytes`

src/test/java/org/cactoos/codec/Base64ToInputTest.java
```diff
@@ -0,0 +1,67 @@
...
+                    new StringReader("SGVsbG8h")
+                )
+            ),
+            new InputHasContent("Hello!")
+        );
+    }
+
+    @Test(expected = DecodingException.class)
+    public void illegalCharacter() throws IOException {
+        final Input decoded = new Base64ToInput(
```
yegor256: @memoyil what is the point of this variable? Can't you just call `stream()` on the value?

yegor256:
@memoyil see more comments above

src/main/java/org/cactoos/codec/Base64ToInput.java
memoyil: If we done like that we could lose information the illegal chars on DecodingException when exists. If we dont want to show illegal chars that the user gave us, we could use >0 . What should we do ?

memoyil: @yegor256 I take that code stub from takes btw :)

src/main/java/org/cactoos/codec/Base64ToInput.java
memoyil: fixed

src/main/java/org/cactoos/codec/InputToBase64.java
memoyil: fixed

src/test/java/org/cactoos/codec/Base64ToInputTest.java
memoyil: fixed

src/main/java/org/cactoos/codec/Base64ToInput.java
memoyil: @yegor256 https://github.com/yegor256/takes/blob/master/src/main/java/org/takes/facets/auth/codecs/CcBase64.java

yegor256: @memoyil I think we should fail if illegal chars are there

memoyil: @yegor256 I have changed the approach

memoyil:
@yegor256 Refactoring the code based on your comments. Thanks in advance

yegor256:
@memoyil I'm really not sure about this strategy. There are many different codec in Apache Codec. Are we going to implement them all? I doubt it. I don't see a lot of value in these classes. We are just wrapping a simple static method, which anyway is coming from JDK. User can do it themselves, just by using lambda. No?

fabriciofx:
@yegor256 @memoyil I think worth implement codecs to some things, for example, `GzipCodec` which has read a buffer to encode/decode. The same to MD5, SHA1 and SHA256 algorithms as you can see in #338 . WDYT?

memoyil:
@yegor256 @fabriciofx I think yegor says no one use this codec implementation as a decorated style or no one needs more than one operation on codecs.Therefor we are just wrapping a static method. Am i right @yegor256?

yegor256:
@memoyil yes, you're right. This functionality is rather simple and used very rarely. It's not so difficult to implement `Input` in-line:

```java
Input origin = ...;
Input input = () -> Base64.getDecoder().decode(BytesOf(origin).bytes());
```

That's it. See?

memoyil:
@yegor256 @fabriciofx So do we cancel the porting codecs action ? WDYT?

yegor256:
@memoyil I think we should cancel that. Maybe we should even add a section to our README explaining which functionality we **don't** implement and why.

memoyil:
@yegor256 All right i am closing that PR then.

0crat:
Oops! Job `gh:yegor256/cactoos#320` was not in scope

PR #323, opened 2017-07-18
Synchronization locks as constructor arguments

g4s8:
We need to be able to use custom lock for synchronization, not always Proc/Func/Scalar as lock.


yegor256:
@rultor merge

yegor256:
@rultor release, tag is `0.11.8`

Issue #324, opened 2017-07-19
AsyncFunc spawns new thread for each apply() call


g4s8:
Creating new threads is expensive operation because it allocates a lot of memory, performs system-calls and register descriptors.
It would be better to create `Executors.newCachedThreadPool()` for async task (or maybe accept Executor as constructor argument) and submit tasks to it:
```java
private final Executor exec = Executors.newCachedThreadPool();

@Override
public Future<Y> apply(final X input) {
    return this.exec.submit(
        () -> this.func.apply(input)
    );
}
```

g4s8:
Looks like it was partially fixed in #330

yegor256:
@0crat in

0crat:
> @0crat in ([here](https://github.com/yegor256/cactoos/issues/324#issuecomment-316656241))

@yegor256 Job `gh:yegor256/cactoos#324` is in scope.

0crat:
Job `gh:yegor256/cactoos#324` assigned to @fabriciofx, please go ahead.

yegor256:
@rultor release, tag is `0.11.7`

yegor256:
@g4s8 the problem seem to be fixed?

g4s8:
@yegor256 sure, thanks

0crat:
Job `gh:yegor256/cactoos#324` is out of scope.

0crat:
+30 points just awarded to @fabriciofx, total is +60

PR #330, opened 2017-07-19
AsyncFunc: executor service as a field

caarlos0:
may avoid the premature termination of the executor service...

### DIFF
--- a/src/main/java/org/cactoos/func/AsyncFunc.java
+++ b/src/main/java/org/cactoos/func/AsyncFunc.java
@@ -23,6 +23,7 @@
  */
 package org.cactoos.func;
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.ThreadFactory;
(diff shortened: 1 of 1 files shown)


src/main/java/org/cactoos/func/AsyncFunc.java
```diff
@@ -91,12 +92,12 @@ public AsyncFunc(final Proc<X> proc, final ThreadFactory fct) {
      */
     public AsyncFunc(final Func<X, Y> fnc, final ThreadFactory fct) {
         this.func = fnc;
-        this.factory = fct;
+        this.executor = Executors.newSingleThreadExecutor(fct);
```
yegor256: @caarlos0 I believe that this ctor must become a secondary one. We should introduce a primary one, which will accept `Func` and `ExecutorService`.

yegor256:
@caarlos0 thanks, see my comment above

caarlos0:
@yegor256 pushed

caarlos0:
this also fixes #324

yegor256:
@rultor merge

caarlos0:
@yegor256 can you deploy it please?

yegor256:
@caarlos0 see 0.11.7

PR #331, opened 2017-07-20
#274: more ctors

yegor256:
see #274

### DIFF
--- a/src/main/java/org/cactoos/func/FuncOf.java
+++ b/src/main/java/org/cactoos/func/FuncOf.java
@@ -95,7 +95,7 @@ public final class FuncOf<X, Y> implements Func<X, Y> {
      * Ctor.
      * @param fnc Func
      */
-    public FuncOf(final Func<X, Y> fnc) {
+    private FuncOf(final Func<X, Y> fnc) {
         this.func = fnc;
     }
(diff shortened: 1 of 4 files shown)


yegor256:
@rultor merge

src/main/java/org/cactoos/func/FuncOf.java
```diff
@@ -95,7 +95,7 @@ public FuncOf(final Proc<X> proc, final Y result) {
      * Ctor.
      * @param fnc Func
      */
-    public FuncOf(final Func<X, Y> fnc) {
+    private FuncOf(final Func<X, Y> fnc) {
```
fabriciofx: @yegor256 I'm sorry, but why make this constructor as private? You are not going to against Cactoos principles, are you?

memoyil: @yegor256 @fabriciofx Yegor is not against private constructors instead he against multiple independent constructors.

fabriciofx: @memoyil Sorry, but I believe private constructors or methods are a code smells. Why? Because they can hide functionality and decrease reusability.

Issue #338, opened 2017-07-22
MD5, SHA-1 and SHA-256 hashes functions


fabriciofx:
Let's implement MD5, SHA-1 and SHA-256 hashes functions. I'll handle it.

0crat:
There is an unrecoverablefailure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.ClassCastException: org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(JsonObjectBuilderImpl.java:252)
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
	at com.zerocracy.radars.github.RbByLabel.react(RbByLabel.java:82)
	at com.zerocracy.radars.github.RbByActions.react(RbByActions.java:65)
	at com.zerocracy.radars.github.Rebound$Chain.react(Rebound.java:74)
	at com.zerocracy.radars.github.RbSafe.lambda$react$0(RbSafe.java:63)
	at org.cactoos.func.FuncWithFallback.apply(FuncWithFallback.java:299)
	at org.cactoos.func.IoCheckedFunc.apply(IoCheckedFunc.java:66)
	at com.zerocracy.radars.github.RbSafe.react(RbSafe.java:79)
	at com.zerocracy.radars.github.RbLogged.react(RbLogged.java:52)
	at com.zerocracy.radars.github.TkGithub.act(TkGithub.java:161)
	at org.takes.facets.fork.FkRegex$1.act(FkRegex.java:154)
	at...
```

0crat:
There is an unrecoverablefailure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.ClassCastException: org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(JsonObjectBuilderImpl.java:252)
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
	at com.zerocracy.radars.github.RbByLabel.react(RbByLabel.java:82)
[...]

Issue #335, opened 2017-07-22
 Bytes to hexadecimal


fabriciofx:
Let's implement a way to represent `Bytes` as a hexadecimal string. I'll handle it.

0crat:
There is an unrecoverablefailure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.ClassCastException: org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(JsonObjectBuilderImpl.java:252)
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
	at com.zerocracy.radars.github.RbByLabel.react(RbByLabel.java:82)
	at com.zerocracy.radars.github.RbByActions.react(RbByActions.java:65)
	at com.zerocracy.radars.github.Rebound$Chain.react(Rebound.java:74)
	at com.zerocracy.radars.github.RbSafe.lambda$react$0(RbSafe.java:63)
	at org.cactoos.func.FuncWithFallback.apply(FuncWithFallback.java:299)
	at org.cactoos.func.IoCheckedFunc.apply(IoCheckedFunc.java:66)
	at com.zerocracy.radars.github.RbSafe.react(RbSafe.java:79)
	at com.zerocracy.radars.github.RbLogged.react(RbLogged.java:52)
	at com.zerocracy.radars.github.TkGithub.act(TkGithub.java:161)
	at org.takes.facets.fork.FkRegex$1.act(FkRegex.java:154)
	at...
```

0crat:
There is an unrecoverablefailure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.ClassCastException: org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl cannot be cast to javax.json.JsonString
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getJsonString(JsonObjectBuilderImpl.java:252)
	at org.glassfish.json.JsonObjectBuilderImpl$JsonObjectImpl.getString(JsonObjectBuilderImpl.java:257)
	at com.zerocracy.radars.github.RbByLabel.react(RbByLabel.java:82)
	at com.zerocracy.radars.github.RbByActions.react(RbByActions.java:65)
[...]

PR #339, opened 2017-07-22
MD5, SHA-1 and SHA-256 hashes functions

fabriciofx:
As per #338. It depends on #335 .

### DIFF
--- a/src/main/java/org/cactoos/io/Md5OfInput.java
+++ b/src/main/java/org/cactoos/io/Md5OfInput.java
@@ -1,0 +1,94 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


memoyil:
@fabriciofx we should put them in the codecs package WDYT?

fabriciofx:
@memoyil I think we should first ask to @yegor256 what he thinks. @yegor256 WDYT? :)

g4s8:
@fabriciofx @memoyil @yegor256 I would not include crypto related code into cactoos library because:
1. This code is platform dependent. Any machine may contain different security provider implementations. Moreover any project (that using cactoos as a dependency) may include third party security provider implementation as a library. The one and the other may change `MessageDigest.getInstance` result unobviously. 
2. It increase complexity by extending the scope from `text+IO+collections` to `text+IO+collections+crypto`

llorllale:
@fabriciofx please close this PR as both issues you [mentioned](https://github.com/yegor256/cactoos/pull/339#issue-131898162) were cancelled by yourself

llorllale:
@fabriciofx closing as per what I [said](https://github.com/yegor256/cactoos/pull/339#issuecomment-382078771)

0crat:

PR #341, opened 2017-07-23
Merged ReplacedText and ReplacedRegexText

ixmanuel:
Reduced the quantity of objects.

### DIFF
--- a/src/main/java/org/cactoos/text/RegexText.java
+++ b/src/main/java/org/cactoos/text/RegexText.java
@@ -1,0 +1,78 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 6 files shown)


ixmanuel:
A better name could be ReplacedOccurrences, but we would have to change the rest of texts: Lowercased, Snakecased, Titlecased. Thus, I let it as ReplacedAllText instead of ReplacedText.

ixmanuel:
@yegor256 A better name could be ReplacedOccurrences, but we would have to change the rest of texts in order to be consistent in the naming: Lowercased, Snakecased, Titlecased. Thus, I let it as ReplacedAllText instead of ReplacedText.

yegor256:
@ixmanuel I don't really understand why do we need `RegexText`? Seems like something redundant to me.

0crat:
Job `gh:yegor256/cactoos#341` is not assigned, can't get start time

0crat:
Job [#341](https://github.com/yegor256/cactoos/issues/341) was not in scope

PR #343, opened 2017-07-24
Added FilteredText

svendiedrichsen:
Added FilteredText implementation. Please review.


src/test/java/org/cactoos/text/FilteredTextTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.cactoos.TextHasString;
+import org.hamcrest.MatcherAssert;
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+/**
+ * Tests for {@link FilteredText}.
+ * @author Sven Diedrichsen (sven.diedrichsen@gmail.com)
+ * @version $Id$
+ * @since 24.07.17
```
fabriciofx: @svendiedrichsen You should use the next cactoos version, in other words, 0.12.

src/main/java/org/cactoos/text/FilteredText.java
```diff
@@ -0,0 +1,76 @@
...
+import org.cactoos.func.IoCheckedFunc;
+
+/**
+ * Text filtered by a {@link org.cactoos.Func} working as a filter.
+ *
+ * @author Sven Diedrichsen (sven.diedrichsen@gmail.com)
+ * @version $Id$
+ * @since 24.07.17
+ */
+public class FilteredText implements Text {
```
fabriciofx: @svendiedrichsen This class should be `final`.

src/test/java/org/cactoos/text/FilteredTextTest.java
svendiedrichsen: I used the master branch to fork from. If thats not correct which one should I have forked from?

src/main/java/org/cactoos/text/FilteredText.java
svendiedrichsen: I'll fix it.

src/main/java/org/cactoos/text/FilteredText.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+package org.cactoos.text;
+
+import java.io.IOException;
+import org.cactoos.Func;
+import org.cactoos.Text;
+import org.cactoos.func.IoCheckedFunc;
+
+/**
+ * Text filtered by a {@link org.cactoos.Func} working as a filter.
```
fabriciofx: @svendiedrichsen I think is more useful filter by strings than by characters. WDYT? Could you show me useful scenarios of filter by character? Thanks in advance!

src/test/java/org/cactoos/text/FilteredTextTest.java
fabriciofx: @svendiedrichsen It's correctm but you should pay attention to the next [release version](https://github.com/yegor256/cactoos/releases).

fabriciofx:
@svendiedrichsen I'm sorry, but have you opened an issue first?

src/test/java/org/cactoos/text/FilteredTextTest.java
svendiedrichsen: I'm sorry but I can't find any info on the release page which tells me that there will be a release 0.12 soon or which branch to use for future PRs.

src/main/java/org/cactoos/text/FilteredText.java
svendiedrichsen: I can't really show you but I can imagine some situations where one would like to filter illegal characters from a text.

svendiedrichsen:
I'm sorry but I haven't. I have just created #344 for this. Sorry.

src/test/java/org/cactoos/text/FilteredTextTest.java
fabriciofx: @svendiedrichsenI agree with you, however we are using as "informal" method look at the release page and put the number of the next expected version and so far it has been working well. :)

fabriciofx:
@svendiedrichsen No problem. Thanks for created the issue!

src/main/java/org/cactoos/text/FilteredText.java
fabriciofx: @svendiedrichsen Well, the only scenarios I can see is to filter whitespaces. Maybe could be more useful filter strings than characters. Or you could filter both: if the string is only one char, you could perform the same thing to filter the character.

svendiedrichsen: Under what circumstances do you need to filter strings? I can't think of any but if you think we need both kind of filtering I think we need two different classes for that. These two are too different for one class imo.

svendiedrichsen:
I have changed FilteredText to use a collection of texts which shall be filtered from the original text.

yegor256:
@svendiedrichsen can you please update our compatibility table in README too? It's important to point out which Apache/JDK/Guava features we're reproducing.

svendiedrichsen:
@yegor256 done.

svendiedrichsen:
I have a little problem with the implementation though. The order in which the filters are removed from the original string matter. So for example:
- You have the string "abcd"
- if you remove the strings "bc", "ad" you end up with an empty string
- if you remove the strings "ad", "bc" you end up with the string "ad"

Either way it is a somewhat suprising result.  I don't know if such an algorithm implementation is such a good idea in the first place. WDYT?

yegor256:
@svendiedrichsen my biggest concern here is the motivation behind this class/feature. I also don't understand why it was introduced in Apache `StringUtils`. What's wrong with `String.replaceAll()`?

svendiedrichsen:
My motivation initially was to have a possibillity to remove illegal characters (Whitespaces, Non-Alpha etc.) from a text. By using a func to determine illegal characters you could pretty much do whatever you liked. @fabriciofx insisted instead to have a list of texts for removal which may be single character. But this introduced the possibility of the above mentioned strange behaviour.

Concerning Apaches `StringUtils.removeAll` motivation: I guess they wanted something NULL safe which `String.replaceAll` isn't.

svendiedrichsen:
Shall I delete this?

svendiedrichsen:
@yegor256 : How should we go here? Close it?

yegor256:
@svendiedrichsen I would close this. Don't see why we need this.

0crat:
I can't do it for technical reasons, I'm very sorry. If you don't know what to do, submit this error as a ticket [here](https://github.com/zerocracy/datum):

```
1.0-SNAPSHOT f2c623f 2017-08-14T16:18:41
IllegalStateException: java.lang.reflect.InvocationTargetException
com.zerocracy.jstk.SoftException: Job `gh:yegor256/cactoos#343` was not in scope
	at com.zerocracy.pm.scope.Wbs.remove(Wbs.java:107)
	at com.zerocracy.pm.scope.Wbs$remove$3.call(Unknown Source)
	at org.codehaus.groovy.runtime.callsite.CallSiteArray.defaultCall(CallSiteArray.java:48)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:113)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.call(AbstractCallSite.java:125)
	at com.zerocracy.stk.pm.scope.wbs.remove_job_from_wbs.exec(remove_job_from_wbs.groovy:32)
	at sun.reflect.GeneratedMethodAccessor54.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.zerocracy.farm.reactive.StkGroovy.process(StkGroovy.java:109)
	at com.zerocracy.farm.StkSafe.process(StkSafe.java:90)
	at com.zerocracy.farm.reactive.Brigade.process(Brigade.java:105)
	at com.zerocracy.farm.reactive.Brigade.pr...
```

PR #345, opened 2017-07-25
#63: Let's do a not procedural temporary file creation/deletion

fabriciofx:
As per #63.


fabriciofx:
@yegor256 Please, could you check if this PR is fine? Thanks in advance!

src/test/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * Create a temporary file used only for tests.
+ *
+ * @author Yegor Bugayenko (yegor256@gmail.com)
+ * @author Fabricio Cabral (fabriciofx@gmail.com)
+ * @version $Id$
+ * @since 0.12
+ */
+public final class TempFile implements Scalar<Path>, Closeable {
```
yegor256: @fabriciofx let's move it to `src/main`. It looks like a good class for other projects.

src/test/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * The Path of the file.
+     */
+    private final Scalar<Path> pth;
+
+    /**
+     * Ctor.
+     */
+    public TempFile() {
```
yegor256: @fabriciofx I would also like to have a ctor with `prefix` only

src/test/java/org/cactoos/io/TempFile.java
```diff
@@ -0,0 +1,86 @@
...
+        this("", "");
```
yegor256: @fabriciofx this won't work, `Files.createTempFile()` will complain because of empty strings. Try to create a unit test for this class and you will see.

yegor256:
@fabriciofx good idea, see my comments above

src/test/java/org/cactoos/io/TempFile.java
fabriciofx: @yegor256 Ok. Good idea.

src/test/java/org/cactoos/io/TempFile.java
fabriciofx: @yegor256 Ok.

src/test/java/org/cactoos/io/TempFile.java
fabriciofx: @yegor256 You're right. I'll make a test for `TestFile` too.

fabriciofx: @yegor256 Update: Apparently this code works (at least in my machine). `new TempFile()` will create a temporary file like `C:\Users\fbc\AppData\Local\Temp\8305850303292415284` (I'm running Windows 10). WDYT? Should I remove this constructor?

llorllale:
@fabriciofx please close this PR; the solution was merged in #556

llorllale:
@fabriciofx closing because #556 solved this

0crat:

PR #355, opened 2017-07-28
LengthOf#274 Text, Input, Iterable and Iterator

ixmanuel:
### [Skip this comment, see next entry please]
This PR has included the PR #354 ArrayOf

### Notes:
- If LenghtOfInput(Input) tries to count single visual characters, then this PR can be used instead of ArrayOf PR #354 and later, this PR should be patched because actually it calculates the length as bytes.length and thus, the length for non-ascii chars doesn't match the length of single visual characters, because non-ascii chars have two or more bytes for their representation.

- If LenghtOf(Input) tries to calculate the size of bytes, then this PR must be updated later with the following considerations: Length for Inputs must be removed from the new LengthOf because semantically it is different and we could consider to rename it to  SizeOfInput(Input).


ixmanuel:
@yegor256 @fabriciofx 

This PR:
- LengthOf(Text) counts the elements.
- LengthOf(Iterable, Iterator) counts the elements.
- LengthOf(Input) the size in bytes.

Thus, it is counterintuitive so, an alternative:
- CountOf  instead of LengthOf to count elements in Text, Bytes, Input, Iterable and Iterator.
- SizeOf to calculate the size in bytes for Text, Bytes and Input

fabriciofx:
@ixmanuel I vote for `CountOf`. @yegor256 WDYT?

ixmanuel:
This PR is outdated and we have to decide. See above.

0crat:
Oops! Job `gh:yegor256/cactoos#355` was not in scope

PR #381, opened 2017-08-05
Fix #376

### LINKED ISSUE #376: RepeatedFunc wrongly throws exception
I expect the following code to run without a problem :
```java
        new RepeatedFunc<>(
            input -> {return null;}, 2
        ).apply(true);
```
Instead I got an exception : `java.lang.IllegalArgumentException: Repeat counter is equal or less than zero: 2`
I don't expect `RepeatedFunc` to behave like `NoNull` implementations, and even if it is the case, the exception message is completely wrong.
The implementation of `RepeatedFunc.apply` should be fixed to check for invalid `this.times` values instead of a `null` result.

ixmanuel:
```java
// IllegalStateException
// The result <null> is an invalid value in Cactoos:
//     - If required, try a null object instead.
new RepeatedFunc<>(
    input -> {return null;}, 2
).apply(true);

// IndexOutOfBoundsException
// Unexpected value=0:
//     - Expected a value > 0 
new RepeatedFunc<>(
    input -> {return 1;}, 0
).apply(true);

```


0crat:
Job `gh:yegor256/cactoos#381` is in scope.

0crat:
+15 points just awarded to @ixmanuel, total is [+75](http://www.0crat.com/u/ixmanuel).

yegor256:
@ixmanuel it looks like over-design. First, we already have `Scalar.NoNulls`. Second, I have big doubts about this new `ValidBound` class -- it looks very artificial and not really Java-ish. Third, the problem you're fixing can be fixed with just 2-3 lines of code.

ixmanuel:
@yegor256, now it is in kiss-mode. Thanks for your feedback in order to be in the same channel.

Some thoughts:
1. Null
```java
// Func (not used because it rises an IO exception)
return new Func.NoNulls<>(this.func).apply(input);

// Scalar (used with an additional statement and new scalar ()->rslt)
final Y rslt = result;
return new Scalar.NoNulls<Y>(()->rslt).value();

// Proposed (deprecated)
return new NoNull(result).value()
```
2. ValidBound's intent is to have uniform-message-exceptions-handling for at least, at most, greater than and less than cases. Not just for this particular case and yes, it is not Java-ish. I'll keep it in mind, thanks.

src/main/java/org/cactoos/func/RepeatedFunc.java
```diff
@@ -75,19 +75,17 @@ public RepeatedFunc(final Func<X, Y> fnc, final int max) {
...
-            throw new IllegalArgumentException(
-                new FormattedText(
-                    "Repeat counter is equal or less than zero: %d",
-                    this.times
-                ).asString()
-            );
-        }
-        return result;
+        final Y rslt = result;
+        return new Scalar.NoNulls<Y>(()->rslt).value();
```
yegor256: @ixmanuel why do we need that? Why can't that function return `null`? I think this is something our users will complain about. This functionality is not what the class is supposed to do.

yegor256:
@ixmanuel see above

ixmanuel:
@yegor256 done :)

yegor256:
@rultor merge

0crat:
Job `gh:yegor256/cactoos#381` assigned to @kkamkou, please go ahead ([policy](http://datum.zerocracy.com/pages/policy.html)).

0crat:
@kkamkou resigned from `gh:yegor256/cactoos#381`, please stop working.

0crat:
Job `gh:yegor256/cactoos#381` assigned to @kkamkou, please go ahead ([policy](http://datum.zerocracy.com/pages/policy.html)).

0crat:
Order was successfully finished: +30 points just awarded to @kkamkou, total is [+930](http://www.0crat.com/u/kkamkou).

Issue #400, opened 2017-08-12
StickyBiFunction


Timmeey:
We already have StickyFunc, lets implement StickyBiFunc

yegor256:
@0crat in

0crat:
> @0crat in ([here](https://github.com/yegor256/cactoos/issues/400#issuecomment-322766042))

@yegor256 Job `gh:yegor256/cactoos#400` is in scope.

memoyil:
@yegor256 You can close this issue

yegor256:
@Timmeey it's yours, what do you think?

Timmeey:
@yegor256 i like the idea of using a map as the "tuple" key. After the minor changes that i requested in the PR are done, this can be closed after merging.

memoyil:
@yegor256 @timmeey Thank you for your comments. As we discussed in pr, no changes need to be done. Thank you

0crat:
Job `gh:yegor256/cactoos#400` is now out of scope.

PR #403, opened 2017-08-17
#400 : Implementing StickBiFunc

memoyil:
Implementing StickBiFunc

### DIFF
--- a/src/main/java/org/cactoos/func/StickyBiFunc.java
+++ b/src/main/java/org/cactoos/func/StickyBiFunc.java
@@ -1,0 +1,78 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.func;
+
+import java.util.HashMap;
+import java.util.Map;
+import org.cactoos.BiFunc;
+import org.cactoos.scalar.StickyScalar;
+
+/**
+ * Func that accepts two arguments and caches previously calculated values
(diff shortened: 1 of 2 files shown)


0crat:
Job `gh:yegor256/cactoos#403` is in scope.

src/main/java/org/cactoos/func/StickyBiFunc.java
```diff
@@ -0,0 +1,78 @@
...
+     * @param fnc Func original
+     */
+    public StickyBiFunc(final BiFunc<X, Y, Z> fnc) {
+        this.func = fnc;
+        this.cache = new HashMap<>(0);
+    }
+
+    @Override
+    public Z apply(final X first, final Y second) throws Exception {
+        final Map<X, Y> keymap = new HashMap<>(1, 1.0F);
```
Timmeey: I did Some digging.

Hashmap calculates a "threshold" for when to resize the map.
We do not want it to resize in this, since the map will only ever contain a max of 1 entry. 

```
/**
     * The next size value at which to resize (capacity * load factor).
    int threshold;
```
```
   /**
   * put method that is used when putting to the map
   */
   final V putVal(int hash, K key, V value, boolean onlyIfAbsent,boolean evict) {
   [...]
        if (++size > threshold)
            resize();
```

This means that with a loadfactor of 1 and the initial capacity the `threshold` will be 1. Therefore putting even ONE entry int the map will trigger a resize operation. This can be avoided here by sacrificing some memory and setting the initial capacity to 2.

memoyil: I corrected it already

Timmeey: But it still says ` final Map<X, Y> keymap = new HashMap<>(1, 1.0F);`? 
1*1.0F is still 1

Oh sorry. I derped. It looks like I read old HashMap code where it said size++>=threshold. but Copy'n'pasted new versions code where it says >. Okay. this can be merged then assume we are all using NON ancient java versions

memoyil: Yes, I think that so. Client must use java8 or higher in order to run this library because of lambda equations. Thank you very much.

memoyil:
@yegor256 We can merge this pr. Thank you

src/main/java/org/cactoos/func/StickyBiFunc.java
```diff
@@ -0,0 +1,78 @@
...
+        keymap.put(first, second);
+        if (!this.cache.containsKey(keymap)) {
+            this.cache.put(keymap, this.func.apply(first, second));
```
yegor256: @memoyil you can use `calculateIfAbsent()`

memoyil: Ok i will fix it

yegor256:
@memoyil since we're in Java 8, let's use all its features, see my comment above

src/main/java/org/cactoos/func/StickyBiFunc.java
memoyil: @yegor256  I wrote that code stub into the function `        final Map<X, Y> keymap = new HashMap<>(1, 1.0F);
        keymap.put(first, second);
        return this.cache.computeIfAbsent(
            keymap,
            tuple -> {
                try {
                    return this.func.apply(first, second);
                } catch (final Exception exp) {
                    throw new RuntimeException(exp);
                }
            }
        );`

But it didnt pass the checkstyle because of catching exception. Since lambda functions could not throw exceptions I am stuck at the moment. What should we do ?

memoyil:
@yegor256 I have encountered some problems as i mentioned above, what are your thoughts about that?

yegor256:
@rultor merge

0crat:
Job `gh:yegor256/cactoos#403` assigned to @kkamkou, please go ahead ([policy](http://datum.zerocracy.com/pages/policy.html)).

0crat:
@kkamkou resigned from `gh:yegor256/cactoos#403`, please stop working.

0crat:
Job `gh:yegor256/cactoos#403` assigned to @kkamkou, please go ahead ([policy](http://datum.zerocracy.com/pages/policy.html)).

0crat:
Order was successfully finished: +30 points just awarded to @kkamkou, total is [+930](http://www.0crat.com/u/kkamkou).

0crat:
Oops! Job `gh:yegor256/cactoos#403` is not in scope

Issue #510, opened 2017-12-31
HexOf


fabriciofx:
Let's implement a way to represent Bytes as a hexadecimal string. I'll handle it.

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.IllegalStateException: BkParallel-6 interrupted while waiting for "roles.xml" in C63314D6Z for 55s
	at com.zerocracy.farm.sync.SyncProject.acq(SyncProject.java:88)
	at com.zerocracy.farm.strict.StrictProject.acq(StrictProject.java:117)
	at com.zerocracy.farm.props.PropsProject.acq(PropsProject.java:78)
	at com.zerocracy.farm.footprint.FtProject.acq(FtProject.java:67)
	at com.zerocracy.farm.uplinked.UplinkedProject.lambda$acq$4(UplinkedProject.java:99)
	at org.cactoos.scalar.Ternary.value(Ternary.java:123)
	at org.cactoos.func.IoCheckedFunc.lambda$apply$0(IoCheckedFunc.java:60)
	at org.cactoos.func.IoCheckedBiFunc.apply(IoCheckedBiFunc.java:66)
	at org.cactoos.func.IoCheckedFunc.apply(IoCheckedFunc.java:61)
	at org.cactoos.scalar.IoCheckedScalar.value(IoCheckedScalar.java:65)
	at com.zerocracy.farm.uplinked.UplinkedProject.acq(UplinkedProject.java:101)
	at com.zerocracy.farm.ruled.RdProject.acq(RdProject.java:54)
	at com.zerocracy.farm.reactive.RvProject.acq(RvProject.java:...
```

[1.0-SNAPSHOT](https://github.com/zerocracy/farm/tree/1.0-SNAPSHOT) /cc @yegor256

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/datum):

```
java.lang.IllegalStateException: BkParallel-0 interrupted while waiting for "roles.xml" in C63314D6Z for 55s
	at com.zerocracy.farm.sync.SyncProject.acq(SyncProject.java:88)
	at com.zerocracy.farm.strict.StrictProject.acq(StrictProject.java:117)
	at com.zerocracy.farm.props.PropsProject.acq(PropsProject.java:78)
	at com.zerocracy.farm.footprint.FtProject.acq(FtProject.java:67)
[...]

PR #511, opened 2017-12-31
#510 HexOf

fabriciofx:
As per #510 .

### DIFF
--- a/src/main/java/org/cactoos/text/HexOf.java
+++ b/src/main/java/org/cactoos/text/HexOf.java
@@ -1,0 +1,81 @@
+/**
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 please, pay attention to this pull request

0crat:
Job `gh:yegor256/cactoos#511` is now in scope.

src/main/java/org/cactoos/text/HexOf.java
```diff
@@ -0,0 +1,81 @@
...
+     * The Bytes.
+     */
+    private final Bytes bytes;
+
+    /**
+     * Ctor.
+     * @param source The bytes
+     */
+    public HexOf(final Bytes source) {
+        this.bytes = source;
```
svendiedrichsen: Wouldn't it be better to use a StickyScalar<Text> internally to cache the result? Here the hex string gets created every time you call asString().

fabriciofx: @svendiedrichsen Maybe could be better a `StickyBytes` instead of alter `HexOf`. @yegor256 What do you think?

yegor256: @fabriciofx I agree, the current implementation is correct.

yegor256:
@rultor merge

0crat:
Oops! Job `gh:yegor256/cactoos#511` is not assigned, can't get start time

0crat:
Job `gh:yegor256/cactoos#511` is now out of scope.

Issue #643, opened 2018-02-14
Calling ItemAt.value() twice doesn't work


aliceice:
If I want to reuse an `ItemAt` it currently throws an IOException.

```
final ItemAt<Integer> item = new ItemAt<>(new IterableOf<>(1));
item.value(); // returns 1
item.value(); // throws IOException
```

I get following Stacktrace:

```
java.io.IOException: The iterable is empty
	at org.cactoos.scalar.ItemAt.lambda$new$0(ItemAt.java:69)
	at org.cactoos.scalar.ItemAt.value(ItemAt.java:253)
```

I'd expect it to return the same object again.

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

0crat:
Job [#643](https://github.com/yegor256/cactoos/issues/643) is now in scope, role is `DEV`

0crat:
The job [#643](https://github.com/yegor256/cactoos/issues/643) assigned to @driver733[/z](https://www.0crat.com/u/driver733), here is [why](https://www.0crat.com/footprint/C63314D6Z/462240122). The budget is 30 minutes, see [§4](http://datum.zerocracy.com/pages/policy.html#4). Please, read [§8](http://datum.zerocracy.com/pages/policy.html#8) and [§9](http://datum.zerocracy.com/pages/policy.html#9). If the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html).

driver733:
@0crat waiting

0crat:
> @0crat waiting ([here](https://github.com/yegor256/cactoos/issues/643#issuecomment-366125663))

@driver733 The impediment for [#643](https://github.com/yegor256/cactoos/issues/643) was registered successfully by @driver733[/z](https://www.0crat.com/u/driver733)

svendiedrichsen:
@aliceice I think you can close this. The PR has been merged.

0crat:
@ypshenychka[/z](https://www.0crat.com/u/ypshenychka) please review this job, as in [§30](http://datum.zerocracy.com/pages/policy.html#30)

ypshenychka:
@0crat Quality good

0crat:
[...]

PR #658, opened 2018-02-16
#643 - ItemAt object now returns correct value when used again

driver733:
As per #643 

I have refactored the class, so that it saves the value at the position specified in the constructor in a private property. When the `value()` method is called again, it simply returns the saved value.

Since we need to save the value inside the object, our class cannot be fully immutable (internally, it is still immutable from outside) as we need to return the value of type T on each method call. I think that using a list with a capacity of 1 element will suffice, although I do agree that this is a compromise.

### DIFF
--- a/src/main/java/org/cactoos/scalar/ItemAt.java
+++ b/src/main/java/org/cactoos/scalar/ItemAt.java
@@ -24,14 +24,16 @@
 package org.cactoos.scalar;
 
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.List;
 import org.cactoos.Func;
 import org.cactoos.Scalar;
 import org.cactoos.text.FormattedText;
 
 /**
  * Element from position in {@link Iterator}
- * or fallback value if iterator hasn't this position.
+ * or fallback saved if iterator hasn't this position.
  *
  * <p>There is no thread-safety guarantee.
  *
@@ -48,7 +50,7 @@ public final class ItemAt<T> implements Scalar<T> {
     private final Scalar<Iterator<T>> src;
 
     /**
-     * Fallback value.
+     * Fallback saved.
      */
     private final Func<Iterable<T>, T> fbk;
 
@@ -56,6 +58,12 @@ public final class ItemAt<T> implements Scalar<T> {
      * Position.
      */
     private final int pos;
+
+    /**
+     * List that holds the value of the iterator
+     *  at the position specified in the constructor.
+     */
+    private final List<T> saved;
 
     /**
      * Ctor.
@@ -74,7 +82,7 @@ public final class ItemAt<T> implements Scalar<T> {
     /**
      * Ctor.
      *
-     * @param fallback Fallback value
+     * @param fallback Fallback saved
      * @param source Iterable
      */
     public ItemAt(final T fallback, final Iterable<T> source) {
@@ -84,7 +92,7 @@ public final class ItemAt<T> implements Scalar<T> {
     /**
      * Ctor.
      *
-     * @param fallback Fallback value
+     * @param fallback Fallback saved
      * @param source Iterable
      */
     public ItemAt(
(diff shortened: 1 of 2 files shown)


0crat:
Job [#658](https://github.com/yegor256/cactoos/issues/658) is now in scope, role is `REV`

amihaiemil:
@driver733 Same here... completed what exactly? Descriptive title and link the parent Issue in the description, please.

driver733:
@amihaiemil Done.

driver733:
@amihaiemil I have decided to use a list, so that the property could be final.

amihaiemil:
@driver733 hmm.. this doesn't seem ok: 

1. it's a strange workaround
2. the class has 2 jobs now: indicating the element at a certain position and also caching it
3. Indeed, we would like all classes to be immutable so all members should be ``final``. 

I believe the solution here is to use a ``CachedScalar`` decorator? Don't we have that in cactoos? 
It would be used something like this (pseudocode):


```java
Scalar itemAt = new CachedScalar (
    new ItemAt(...)
)
```

And it makes sense -- an iterator will pass over an element exactly once, then it will eventuall reach its end and stop, it doesn't go back. Also when you going over an iterator, if you want to keep any value, you have to cache it. Same is here. 

Since you already worked a lot here, now it's time to cut corners! I would leave a Javadoc on ``ItemAt``, explaining that it's a one-time operation and if they want to cache that value, they have to use a caching decorator. Then, leave a puzzle for implementing such a caching decorator.

driver733:
@amihaiemil I absolutely agree with you. However, I think that it would be better, in such case, to refactor this class completely or to get rid of it. I think that the class named as `ItemAs` is responsible for returning the same value each time, without asking the user to use a caching decorator. Actually, a better workaround would be to use a `@Cachable` annotation from `jacbi-aspects`.
What do you think?

svendiedrichsen:
@driver733 @amihaiemil IMHO the only member the `ItemAt` class needs is a `StickyScalar<T>` which should be created in the ctor. It can evaluate the item using the code currently in `value()` method and will cache the result. WDYT?

driver733:
@svendiedrichsen So the `value()` method would be called in the constructor and then would be saved in a `StickyScalar<T>` private property?

svendiedrichsen:
@driver733 No, I mean that in the Ctor. you create a `StyickScalar<T>` with a body which consists of the code which is currently in the `value()` method. Thus you only have the scalar as member to the `ItemAt` class and in the `value()` method you call `this.scalar.value()` everytime.

svendiedrichsen:
@driver733 You may have to wrap the `StickyScalar` in a `UncheckScalar` though.

driver733:
@svendiedrichsen I like you suggestion.
@amihaiemil I believe that we should refactor this class in that way. Do you agree?

svendiedrichsen:
@driver733 Looks good.

amihaiemil:
@driver733 Yes, looks much more elegant this way (@svendiedrichsen thank you for the suggestion). 

And yes, ``@Cachable`` would also be a nice solution, but I don't know if cactoos already uses jcabi-aspects or not and if Yegour would like to use it just for this. 

This being said, I think you reached a nice solution :)

amihaiemil:
@rultor good to mege

amihaiemil:
@yegor256 ping

yegor256:
@rultor merge

0crat:
@ypshenychka[/z](https://www.0crat.com/u/ypshenychka) please review this job, as in [§30](http://datum.zerocracy.com/pages/policy.html#30)

ypshenychka:
@0crat Quality good

0crat:
QA review completed: +15 points just awarded to @ypshenychka[/z](https://www.0crat.com/u/ypshenychka), total is [+225](https://www.0crat.com/u/ypshenychka)

0crat:
Order was finished, quality was "good": +20 points just awarded to @amihaiemil[/z](https://www.0crat.com/u/amihaiemil), total is [+715](https://www.0crat.com/u/amihaiemil)

0crat:
Payment to `ARC` for a closed pull request, as in [§28](http://datum.zerocracy.com/pages/policy.html#28): +15 points just awarded to @yegor256[/z](https://www.0crat.com/u/yegor256), total is [+10259](https://www.0crat.com/u/yegor256)

Issue #711, opened 2018-02-22
NormalizedText.asString() is too procedural.


aliceice:
The method `asString()` of `NormalizedText` currently looks like this:

```
@Override
public String asString() throws Exception {
    return new TrimmedText(this.origin).asString().replaceAll("\\s+", " ");
}
```

Instead of calling `.replaceAll(...)` it should use `ReplacedText` to build a bigger object and make it more declarative.

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

svendiedrichsen:
@yegor256 May I contribute to this?

memoyil:
@aliceice you can check the pr.

memoyil:
@0crat assign

0crat:
You need to ask somebody to invite you: https://github.com/zerocracy/datum#invite-only

llorllale:
@aliceice #743 was merged, can we close this?

llorllale:
@aliceice can we close this issue?

llorllale:
@aliceice I'm closing this

0crat:
The job is not in WBS, won't close the order

PR #739, opened 2018-03-14
#711: Change normalized string asstring with ReplacedText

memoyil:
Change normalized string asstring with ReplacedText


0crat:

Issue #747, opened 2018-03-23
Reduced and Folded are confused


VsSekorin:
Now `org.cactoos.scalar.Folded` is reduce and `org.cactoos.scalar.Reduced` is fold.
You can check it in Scala or another programming language. You can also see [cactoos-math FoldLeft](https://github.com/VsSekorin/cactoos-math/blob/master/src/main/java/com/vssekorin/cactoosmath/scalar/FoldLeft.java) and  [cactoos-math ReduceLeft](https://github.com/VsSekorin/cactoos-math/blob/master/src/main/java/com/vssekorin/cactoosmath/scalar/ReduceLeft.java).

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

0crat:
Job [#747](https://github.com/yegor256/cactoos/issues/747) is now in scope, role is `DEV`

0crat:
Bug was reported, see [§29](http://www.zerocracy.com/policy.html#29): +15 point(s) just awarded to @vssekorin[/z](https://www.0crat.com/u/vssekorin)

SharplEr:
@0crat status

0crat:
> @0crat status ([here](https://github.com/yegor256/cactoos/issues/747#issuecomment-383283160))

@SharplEr This is what I know about this job, as in [§32](http://www.zerocracy.com/policy.html#32):

  * The job [#747](https://github.com/yegor256/cactoos/issues/747) is [in scope](http://www.zerocracy.com/policy.html#14) for 9days
  * The role is `DEV`
  * The job is not assigned to anyone
  * The [budget](http://www.zerocracy.com/policy.html#4) is not set yet
  * These users are banned and won't be assigned:
    * @vssekorin[/z](https://www.0crat.com/u/vssekorin): This user reported the ticket
  * Job [footprint](https://www.0crat.com/footprint/C63314D6Z?q=%7Bjob%3A%27gh%3Ayegor256%2Fcactoos%23747%27%2Ctype%3A%7B%24not%3A%2F%5ENotify%2F%7D%7D) (restricted area)

0crat:
[...]

PR #798, opened 2018-04-30
Resolution of #747, "Reduced and Folded Confused"

paulodamaso:
Resolution of #747, which states that 'Reduced' and 'Folded' are inverted. Swapped 'Reduced' and 'Folded' classes and test cases.

### DIFF
--- a/src/main/java/org/cactoos/scalar/Folded.java
+++ b/src/main/java/org/cactoos/scalar/Folded.java
@@ -23,108 +23,68 @@
  */
 package org.cactoos.scalar;
 
-import java.util.Iterator;
-import java.util.NoSuchElementException;
 import org.cactoos.BiFunc;
 import org.cactoos.Scalar;
-import org.cactoos.iterable.Mapped;
+import org.cactoos.iterable.IterableOf;
 
 /**
- * Folds iterable via BiFunc.
+ * Iterable, which elements are "folded" through the func.
  *
- * <pre>
- * new Folded<>(
- *     (first, last) -> first + last,
- *     new IterableOf<>(() -> 1L, () -> 2L, () -> 3L, () -> 4L)
- * ).value() // returns 10L
- * </pre>
- *
- * <p>Here is how you can use it to
- * find one of items according to the specified {@link BiFunc}:</p>
- *
- * <pre>
- *  final String apple = new Folded&lt;&gt;(
- *          (first, last) -&gt; first,
- *          new IterableOf&lt;Scalar&lt;String&gt;&gt;(
- *              () -&gt; "Apple",
- *              () -&gt; "Banana",
- *              () -&gt; "Orange"
- *          )
- *      ).value();
- *  final String orange = new Folded&lt;&gt;(
- *          (first, last) -&gt; last,
- *          new IterableOf&lt;Scalar&lt;String&gt;&gt;(
- *              () -&gt; "Apple",
- *              () -&gt; "Banana",
- *              () -&gt; "Orange"
- *          )
- *      ).value();
- * </pre>
- *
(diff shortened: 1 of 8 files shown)


0crat:
Job [#798](https://github.com/yegor256/cactoos/issues/798) is now in scope, role is `REV`

src/main/java/org/cactoos/scalar/Folded.java
```diff
@@ -23,108 +23,68 @@
...
- *
- * @author Alexander Dyadyushenko (gookven@gmail.com)
- * @author Eduard Balovnev (bedward70@mail.ru)
+ * @author Yegor Bugayenko (yegor256@gmail.com)
  * @version $Id$
- * @param <T> Scalar type
- * @since 0.29
+ * @param <T> Type of element
+ * @param <X> Type of input and output
+ * @since 0.30
```
proshin-roman: @paulodamaso I'm afraid if changing the `@since` tag is a good idea. Classes and their functionality existed before and you just changed their names/content. I think it would be better to keep the versions of the functionality: Folded -> since  0.9, Reduced -> since 0.29. WDYT?

paulodamaso: @proshin-roman I'm thinking about naming, authoring and functionality/content: the file `Folded.java`, whose author is Yegor and which represents an "iterable which elements are 'folded' througn the function" appeared just in version 0.30, not before; the same with `Reduced.java`: we have not this name-author-content configuration (`Reduced.java`, which "Reduces iterable by BiFunc", by Alexander and Eduard) in version 0.29. Keeping the old version does not seems right to me because we would not respect these authoring information against name, version and content of each class.

proshin-roman: @paulodamaso To be honest, I don't think that authoring is the most important info in the code :) In any case, the code is changing constantly by many developers and at some point the class contains nothing written by the mentioned author. What is more important IMO is a relation of name, version, and content. Let's say, somebody uses the library of version 0.5 (for example). Then he can look at Folded class and find the version when this class has appeared the first time (0.29). And then he can just download this specific 0.29 version and then this file MUST be in the code. The problem that the class in version 0.29 has absolutely different functionality comparing to the same file in version 0.30. 

So... yes, I agree that we should specify the current version (0.30) - just because it's safer.

proshin-roman:
@llorllale this pull request can be merged (I mention you because rultor doesn't react to my message).

llorllale:
@rultor merge

0crat:
@elenavolokhova[/z](https://www.0crat.com/u/elenavolokhova) please review this job completed by @proshin-roman[/z](https://www.0crat.com/u/proshin-roman), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
The job [#798](https://github.com/yegor256/cactoos/issues/798) is now out of scope

0crat:
Payment to `ARC` for a closed pull request, as in [§28](http://www.zerocracy.com/policy.html#28): +10 point(s) just awarded to @llorllale[/z](https://www.0crat.com/u/llorllale)

elenavolokhova:
@proshin-roman According to our Policy:
> The code reviewer found at least three problems in the code

Please, confirm that you'll try to find at least 3 issues in future reviews.

proshin-roman:
> Please, confirm that you'll try to find at least 3 issues in future reviews.

@elenavolokhova confirm

0crat:
Order was finished, quality is "acceptable": +15 point(s) just awarded to @proshin-roman[/z](https://www.0crat.com/u/proshin-roman)

Issue #776, opened 2018-04-17
improve reliability of LoggingInputTest


ashtonhogan:
I keep seeing this [error](https://ci.appveyor.com/project/yegor256/cactoos/build/1113) even though it has nothing to do with what I'm changing. I also don't have sufficient permission to trigger rebuild myself so this results in blocking PR's.

```
Tests run: 10, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.047 sec <<< FAILURE! - in org.cactoos.io.LoggingInputTest
logIntoCreatedLogger(org.cactoos.io.LoggingInputTest)  Time elapsed: 0.047 sec  <<< FAILURE!
java.lang.AssertionError: 
Expected: a string containing "Read 8 byte(s)"
     but: was ""
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.cactoos.io.LoggingInputTest.logIntoCreatedLogger(LoggingInputTest.java:231)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.apache.maven.surefire.junitcore.pc.Scheduler$1.run(Scheduler.java:393)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[...]

PR #855, opened 2018-05-10
Fix LoggingInputTest errors in appveyor

victornoel:
This is for #776.

Basically, as the `Logger.getLogger()` method warns, if you don't retain a strong reference to a logger retrieved from this method, nothing prevent its garbage collection.
In this case, it meant the `Handler` we added was lost and so the test failed sometimes.

The solution:
- keep a (strong) reference to the logger during the whole test.
- use this reference to remove the handler.
- do NOT rely on the logger itself to create `LoggingInput` though (as this is the purpose of the test).

### DIFF
--- a/src/test/java/org/cactoos/io/LoggingInputTest.java
+++ b/src/test/java/org/cactoos/io/LoggingInputTest.java
@@ -219,13 +219,14 @@ public final class LoggingInputTest {
     @Test
     public void logIntoCreatedLogger() {
         final FakeHandler handler = new FakeHandler();
-        final String logger = "my source";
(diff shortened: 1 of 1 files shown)


0crat:
@victornoel[/z](https://www.0crat.com/u/victornoel) this pull request is too small, just 9 lines changed (less than 10), there will be no formal code review; in the future, try to make sure your pull requests are not too small; @llorllale[/z](https://www.0crat.com/u/llorllale) please review this and merge or reject

llorllale:
@rultor merge

Issue #881, opened 2018-05-18
CollectionEnvelope.java:39-42: Implement methods equals...


0pdd:
The puzzle `844-c2e67e6a` from #844 has to be resolved:

https://github.com/yegor256/cactoos/blob/7775a67e7a6ce47f8a6a54309a21221147cb0328/src/main/java/org/cactoos/collection/CollectionEnvelope.java#L39-L42

The puzzle was created by Roman Proshin on 13-May-18. 

Estimate: 30 minutes,     

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
Job [#881](https://github.com/yegor256/cactoos/issues/881) is now in scope, role is `DEV`

0crat:
The job [#881](https://github.com/yegor256/cactoos/issues/881) assigned to @vatavuk[/z](https://www.0crat.com/u/vatavuk), here is [why](https://www.0crat.com/footprint/C63314D6Z/1381569500); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html)

Vatavuk:
@0crat wait for reviewer

0crat:
> @0crat wait for reviewer ([here](https://github.com/yegor256/cactoos/issues/881#issuecomment-390954000))

@Vatavuk The impediment for [#881](https://github.com/yegor256/cactoos/issues/881) was registered successfully by @vatavuk[/z](https://www.0crat.com/u/vatavuk)

0pdd:
The puzzle `844-c2e67e6a` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd the puzzle [#904](https://github.com/yegor256/cactoos/issues/904) is still not solved.

[...]

PR #894, opened 2018-05-20
#881 equals and hashCode for CollectionEnvelope

Vatavuk:
Introduced `equals` and `hashCode` in `CollectionEnvelope` for #881

### DIFF
--- a/src/main/java/org/cactoos/collection/CollectionEnvelope.java
+++ b/src/main/java/org/cactoos/collection/CollectionEnvelope.java
@@ -30,7 +30,6 @@ import org.cactoos.iterator.Immutable;
 import org.cactoos.scalar.And;
 import org.cactoos.scalar.Folded;
 import org.cactoos.scalar.InheritanceLevel;
-import org.cactoos.scalar.Or;
 import org.cactoos.scalar.SumOfIntScalar;
 import org.cactoos.scalar.UncheckedScalar;
 
@@ -151,14 +150,10 @@ public abstract class CollectionEnvelope<X> implements Collection<X> {
     public final boolean equals(final Object other) {
         return new UncheckedScalar<>(
             new And(
-                new Or(
-                    () -> new InheritanceLevel(
-                        other.getClass(), Collection.class
-                    ).value() > -1,
-                    () -> new InheritanceLevel(
-                        other.getClass(), CollectionEnvelope.class
-                    ).value() > -1
-                ),
+                () -> other != null,
+                () -> new InheritanceLevel(
+                    other.getClass(), CollectionEnvelope.class
+                ).value() > -1,
                 () -> {
                     final Collection<?> compared = (Collection<?>) other;
                     return this.size() == compared.size();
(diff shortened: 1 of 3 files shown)


0crat:
Job [#894](https://github.com/yegor256/cactoos/issues/894) is now in scope, role is `REV`

src/main/java/org/cactoos/collection/CollectionEnvelope.java
```diff
@@ -147,4 +147,48 @@ public String toString() {
...
+    @Override
+    public final boolean equals(final Object other) {
+        return new UncheckedScalar<>(
+            new And(
+                new Or(
+                    () -> new InheritanceLevel(
+                        other.getClass(), Collection.class
+                    ).value() > -1,
+                    () -> new InheritanceLevel(
+                        other.getClass(), CollectionEnvelope.class
```
llorllale: @Vatavuk I believe this isn't necessary because `CollectionEnvelope` *is a* `Collection` already

src/main/java/org/cactoos/collection/CollectionEnvelope.java
```diff
@@ -147,4 +147,48 @@ public String toString() {
         return this.col.value().toString();
     }
 
+    @Override
+    public final boolean equals(final Object other) {
```
llorllale: @Vatavuk this `equals` is missing the null-check

llorllale:
@Vatavuk left you a couple of comments, see above.

Great PR.

src/main/java/org/cactoos/collection/CollectionEnvelope.java
Vatavuk: @llorllale fixed in d4538f8

Vatavuk:
@llorllale fixed and added a comment regarding `InheritanceLevel`

src/main/java/org/cactoos/collection/CollectionEnvelope.java
Vatavuk: @llorllale the end result is that we will compare only classes that are related to `CollectionEnvelope`

llorllale: @Vatavuk We should be comparing `Collection`s here, not `CollectionEnvelope`s.

I think I understand the problem now: `InheritanceLevel` does not check if a class implements an interface (maybe it shouldn't).

We're missing tests that compare `CollectionEnvelope` against regular JDK `Collection`s. I believe they will fail with the current implementation. So we either need to a) add an `instanceof` (against our principles) or b) build a new type that checks if a class implements/extends an interface, or c) modify `InheritanceLevel` to take into account (b).

Please leave a puzzle for this and we'll merge.

Vatavuk:
@llorllale added a puzzle

llorllale:
@Vatavuk there's a problem with the puzzle's format

Vatavuk:
@llorllale fixed

0crat:
The job [#894](https://github.com/yegor256/cactoos/issues/894) is now out of scope

PR #744, opened 2018-03-20
#704 Enforce failure reasons in tests

neonailol:
Implement #704 as suggested in #588

Issue #704, opened 2018-02-21
Enforce failure reasons in tests


driver733:
I think that it would be great if we could enforce the usage of failure reasons in tests. If a test uses a test method that can have a failure reason and it does not use it,  then the build should fail.

0crat:
@yegor256[/z](https://www.0crat.com/u/yegor256) please, pay attention to this issue

yegor256:
@driver733 how is it technically possible?

driver733:
@yegor256 We could, for instance, ban this method
`<T> void assertThat(T actual, Matcher<? super T> matcher) `
then, you would be directed to use one of these methods instead
`<T> void assertThat(java.lang.String reason, boolean assertion) `
`<T> void assertThat(java.lang.String reason, T actual, Matcher<? super T> matcher) `

yegor256:
@driver733 "ban" how?

driver733:
@yegor256 Could we use Checkstyle's [RegexpMultiline](http://checkstyle.sourceforge.net/config_regexp.html#RegexpSinglelineJava#RegexpMultiline)?

yegor256:
@driver733 I doubt it will work... can you try?

driver733:
@yegor256 Sure, I will see what I can do.

driver733:
@yegor256 Actually, we could use this
```
<module name="com.github.sevntu.checkstyle.checks.coding.ForbidCertainMethodCheck">
      <property name="methodName" value="assert(True|False)"/>
      <property name="argumentCount" value="1"/>
    </module>
```

yegor256:
@driver733 yes, that will work, thanks

0crat:
Job [#704](https://github.com/yegor256/cactoos/issues/704) is now in scope, role is `DEV`

0crat:
Bug was reported, see [§29](http://datum.zerocracy.com/pages/policy.html#29): +15 points just awarded to @driver733[/z](https://www.0crat.com/u/driver733), total is [+335](https://www.0crat.com/u/driver733)

0crat:
[...]

PR #882, opened 2018-05-19
Tests are more OO 

victornoel:
This is the first step toward more OO tests #588

- Introduce everything needed in Maven to forbid calls to static `Matchers`: for now it does not fail the build.
- Add `@todo` for continuing the work until everything is converted to pure OO
- This also ignore a test and add a `@todo` for a bug found in `CollectionOf`

PR #899, opened 2018-05-22
#704 - Enforce failure reasons in tests 

neonailol:
Same as #744 but only prints errors without failing the build


0crat:
Job [#899](https://github.com/yegor256/cactoos/issues/899) is now in scope, role is `REV`

llorllale:
@neonailol please use the maven configuration merged in #882.

neonailol:
@llorllale updated

pom.xml
```diff
@@ -156,7 +156,7 @@ The MIT License (MIT)
                -->
               <failOnViolation>false</failOnViolation>
               <signaturesFiles>
-                <signaturesFile>./src/test/resources/forbidden-apis.txt</signaturesFile>
+                <signaturesFile>./src/verifier/forbiddenapis.txt</signaturesFile>
```
llorllale: @neonailol why not use `./src/test/resources/forbidden-apis.txt` like in #882

llorllale:
@neonailol see my comment above

pom.xml
neonailol: Because it is not part of sources, just like verifications.xml

llorllale: @neonailol I don't understand. `src/test/resources` is exactly the path for these kinds of things. It won't be exported to production when released.

neonailol: @llorllale well `src/test/resources` as stated from the path, is the directory for resources used by tests, this, on the other hand, is the resources needed to perform static verification of all sources. I expect in the future someone will be using this to prohibit uses of some other methods in library code, not just the tests.

llorllale: @neonailol we'll do that when the need arises. For now, please put the file back in `src/test/resources`, thanks

neonailol:
@llorllale done, it is also worth to check out bundled with the plugin [signatures](https://github.com/policeman-tools/forbidden-apis/wiki/BundledSignatures)

llorllale:
@neonailol good to know, thanks

llorllale:
@rultor merge

0crat:
The job [#899](https://github.com/yegor256/cactoos/issues/899) is now out of scope

PR #959, opened 2018-10-22
# 955 Refactor text envelope equals

strokyl:
I did not handle null in the equals method because it wasn't handle before. I am new to this project and to this way of coding, even if I road your two books, I am a bit disoriented.

I am a bit puzzle by the fact your linter does not accept variable namedLikeThat, would be happy to know why ?

Because previous code handle equals with object not being an instance of Text but there were no test, I added one.

About my Javadoc, I know it's not good, but I am not used to Javadoc my method especially when there are private (I think I name them and type them in a way they often don't need Javadoc).

Also, do you have a suggestion for a plugin for IDEA to run all your linter simply?

### DIFF
--- a/src/main/java/org/cactoos/text/TextEnvelope.java
+++ b/src/main/java/org/cactoos/text/TextEnvelope.java
@@ -26,7 +26,9 @@ package org.cactoos.text;
 import java.io.IOException;
 import org.cactoos.Scalar;
 import org.cactoos.Text;
+import org.cactoos.scalar.And;
 import org.cactoos.scalar.IoCheckedScalar;
+import org.cactoos.scalar.Or;
 import org.cactoos.scalar.UncheckedScalar;
 
 /**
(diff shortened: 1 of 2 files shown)


0crat:
Job [#959](https://github.com/yegor256/cactoos/issues/959) is now in scope, role is `REV`

src/main/java/org/cactoos/text/TextEnvelope.java
```diff
@@ -73,23 +75,36 @@ public final int hashCode() {
...
-        if (!(obj instanceof Text)) {
-            return false;
-        }
-        final Text that = (Text) obj;
-        return new UncheckedText(this).asString().equals(
-            new UncheckedText(that).asString()
+        return new UncheckedScalar<>(
+            new Or(
+                () -> this == obj,
+                this.equalsOther(obj)
```
llorllale: @strokyl any reason why you split this over three methods?

llorllale:
@strokyl 

> I did not handle null in the equals method because it wasn't handle before.

Can you add that as well please?

> your linter does not accept variable namedLikeThat, would be happy to know why ?

See Yegor's [post](https://www.yegor256.com/2015/01/12/compound-name-is-code-smell.html).

src/main/java/org/cactoos/text/TextEnvelope.java
llorllale: @strokyl please leave it all inside the `equals` method

* I found no explicit mention of this by Yegor, but I believe he also frowns upon compound names for methods (so do I). If you look at the code produced in his other projects you'll see how this idea is implicitly enforced throughout.
* Personally I find these extra methods unnecessary - the implementation required here isn't too complex to justify them

strokyl: Ok I will rework that tonight. Thank you for your feedback

strokyl: done

llorllale:
@rultor merge

llorllale:
@strokyl thanks!

0crat:
Job `gh:yegor256/cactoos#959` is not assigned, can't get performer

0crat:
The job [#959](https://github.com/yegor256/cactoos/issues/959) is now out of scope

PR #960, opened 2018-10-25
944: allow DateOf to parse date without time

strokyl:


0crat:
Job [#960](https://github.com/yegor256/cactoos/issues/960) is now in scope, role is `REV`

src/main/java/org/cactoos/time/DateOf.java
```diff
@@ -64,13 +68,29 @@ public DateOf(final CharSequence date, final String format) {
...
-                    formatter.parse(date)
-                ).toInstant(ZoneOffset.UTC)
-            )
-        );
+            new Ternary<>(
+                () -> temporal.isSupported(ChronoField.HOUR_OF_DAY),
+                () -> Date.from(
+                    LocalDateTime
+                        .from(temporal)
+                        .toInstant(ZoneOffset.UTC)
```
llorllale: @strokyl this temporal accessor may already come with its own zone offset

llorllale:
@strokyl `cactoos` is still in alpha, so we're free to change things. You're right - patterns for `SimpleDateFormat` and `DateTimeFormatter` are different, but this is one of those instances where I'm OK with breaking things.

Also, please add description to this PR. :)

src/main/java/org/cactoos/time/DateOf.java
strokyl: Well seen, I added test for datetime with timezone and took the timezone of the temporal if it has any

strokyl:
`cactoos is still in alpha, so we're free to change things. You're right - patterns for SimpleDateFormat and DateTimeFormatter are different, but this is one of those instances where I'm OK with breaking things.`

Yes, we could break the API and use SimpleDateFormat, but I do not think it would be a good idea because all other class about Date (LocalDateTimeOf, ZonedDateTimeOf) use DateTimeFormatter for parsing. As a user of this library, I would expect all class about date to use same pattern syntax.

Also, to give my opinion, I think Date class from Java 7 is bad, and should be used only for legacy reason. I wouldn't expect a recent library to provide me any direct formatting to Date.
I have been worked a lot with DateTime in a previous job, and I know it's easy to make big mistakes with date and datetime because you never think enough about what you really want:

- a date without a timezone: when you don't care about the real moment but about the hour and minute of the day no matter the place in the word (LocalDateTime). Example, when you say I go to bed at 11pm, you go to bed at 11pm at NewYork or Paris even if this is not the same exact moment.

- a specific moment in time. You don't really care about the timezone but you want to make sure this represent the time anywhere in the world (Instant)

- a day without a time, in this case using DateTime with time at midnight would be a mistake because considering timezone it could become an other day...

- a date with a timezone (ZonedDateTime). In the pass, I thought timezone does not really matter (expect at parsing and showing datetime to final user), and that it was more simple to convert all TimeZoned date in UTC. But sometime, it's a mistake. For example, the datetime of a recurring event should be store with it timezone, otherwise if you event repeat all day, you will have some surprise with DST...

For those reason, I think new class about Date in Java8 are really good because they force the developer to think about all of that. And we should allow the developer to create Java 7 Date only from instant. In the other hand, we should allow him to parse to:
- Instant (failing if no TZ neither time)
- LocalDateTime (failing if TZ or if no time provided)
- ZonedDateTime(failing if no TZ or no time provided)
- LocalDate (failing if TZ or time provided)
If we make a function that make hidden choice about what to do when no timezone, time and converting to Date. The comportment of it, would not be easy to understand for API user.

In the case where the API user would need to parse datetime that can or can not have time or TZ.
I think it would be great to return a wrapper around Temporal to simplify the use of it, or at least return a Temporal from it.

So to be honest, DateOf should have only one ctor that would only accept Instant. And we should InstantOf, LocalDateOf and TemporalAccessOf (or something similar).

llorllale:
@strokyl that's quite a long comment :(

Are you suggesting we remove `DateOf`?

llorllale:
@strokyl ping

strokyl:
No I am suggesting to make DateOf having only one ctor which would take a Instant

llorllale:
@strokyl what would `TemporalAccessOf` be for?

strokyl:
Not sure if this is necessary in a first time.
But imagine you want to parse in a generic way time/datetime and have different action depending, if the string your parsed contain timezone, time etc.. etc...
You will therefore need TemporalAccess or something similar

llorllale:
@strokyl shall we close this PR then?

llorllale:
@strokyl ping

strokyl:
Ok

Le sam. 8 déc. 2018 à 02:05, George Aristy <notifications@github.com> a
écrit :

Issue #911, opened 2018-06-20
Substitute all usages of Iterator with usage of Iterable in ctors


svendiedrichsen:
The discussion from #906 lead to the conclusion to avoid the usage of Iterators and replace all of their occurrences in ctors with Iterables where applicable. This aspect has also been discussed in #677 .

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
There is an unrecoverable failure on my side. Please, submit it [here](https://github.com/zerocracy/farm/issues):

```
PID: 4@699dad38-11e0-4927-b0e8-65c95cb08d43, thread: pool-35-thread-1
com.zerocracy.farm.sync.SyncProject[70] java.lang.IllegalStateException: Failed to acquire "catalog.xml" in "PMO" in 2min: 266bf5224b6f/32min/0/2/true/false by AsyncFlush-2
```

[0.22.16](https://github.com/zerocracy/farm): Issue: yegor256/cactoos#911, Action: `opened`

llorllale:
@0crat in

0crat:
> @0crat in ([here](https://github.com/yegor256/cactoos/issues/911#issuecomment-399082089))

@llorllale Job [#911](https://github.com/yegor256/cactoos/issues/911) is now in scope, role is `DEV`

0crat:
Bug was reported, see [§29](http://www.zerocracy.com/policy.html#29): +15 point(s) just awarded to @svendiedrichsen[/z](https://www.0crat.com/u/svendiedrichsen)

0crat:
The job [#911](https://github.com/yegor256/cactoos/issues/911) assigned to @verkhovin[/z](https://www.0crat.com/u/verkhovin), here is [why](https://www.0crat.com/footprint/C63314D6Z/8c0ad291-2db2-48f2-b939-59ee953120a2); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

verkhovin:
[...]

PR #961, opened 2018-10-28
Correlated iterator: Check a relation between elements of two iterators

dgroup:


0crat:
Job [#961](https://github.com/yegor256/cactoos/issues/961) is now in scope, role is `REV`

llorllale:
@dgroup I don't like the semantics of this, but before we get into that I'd like to ask: what's the use case for this?

dgroup:
Hi @llorllale, as alternative we can rename to `Duplex`.
```java
package com.github.dgroup.dockertest.hamcrest;

import com.github.dgroup.dockertest.iterator.Duplex;
import java.util.Collection;
import org.cactoos.Scalar;
import org.cactoos.iterator.LengthOf;
import org.cactoos.list.ListOf;

/**
 * Hamcrest matcher to compare two collections of {@link String} and ensure that
 *  the elements from collection A ends with elements from collection B.
 *
 * @author Yurii Dubinka (yurii.dubinka@gmail.com)
 * @version $Id: 552e8db400901629271408013f265e829bc9611b $
 * @since 1.0
 */
public final class ItemsEndsWith extends HamcrestCollectionEnvelope<String> {

    /**
     * Ctor.
     * @param expected The expected ending of the actual elements.
     */
    public ItemsEndsWith(final String... expected) {
        this(() -> new ListOf<>(expected));
    }

    /**
     * Ctor.
     * @param expected The expected ending of the actual elements.
     * @checkstyle IndentationCheck (10 lines)
     */
    public ItemsEndsWith(final Scalar<Collection<String>> expected) {
        super(
            (exp, act) -> new LengthOf(
                new Correlated<>(
                    (eval, aval) -> aval.endsWith(eval),
                    exp, act
                )
            ).value() > 0,
            expected
        );
    }

}

``` 
I don't want to publish a new library to maven central with collection/iterator primitives and belive that this functionality follow EO principles.

dgroup:
Hi @llorllale ,
Is there any review comments regarding this RR?

llorllale:
@dgroup looks like what you want is a `Scalar<Boolean>` not an iterator.

dgroup:
@llorllale , no, in example above i'm using `BiFunc`
```java
    public HamcrestCollectionEnvelope(
        final BiFunc<Collection<X>, Collection<X>, Boolean> fnc,
        final Scalar<Collection<X>> exp
    ) {
        this.fnc = new UncheckedBiFunc<>(fnc);
        this.expected = new UncheckedScalar<>(exp);
    }
```
The key point of this is that i need iterator which allows me to iterate over 2 iterators/iterables/collections and verify that elements with the same position has correlation by function. The function might be `equals`, `endsWith`,  `greaterThen`, `nonNull`, `emtpy`, etc,

llorllale:
@dgroup thanks for your explanation. I was actually not really understanding this at first. :)

Do you think `Matched` would be a better name?

src/main/java/org/cactoos/iterator/Correlated.java
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Correlated iterator.
+ *
+ * Iterates over 2 iterators at the same time and verify the correlation
+ *  between elements.
+ *
+ * @param <X> Type of item.
+ * @since 0.39
+ */
+public final class Correlated<X> implements Iterator<X> {
```
llorllale: @dgroup why not `Iterable`? We're trying to avoid iterators - see #911

src/main/java/org/cactoos/iterator/Correlated.java
```diff
@@ -0,0 +1,117 @@
...
+        final Iterator<X> snd
+    ) {
+        this.fnc = new UncheckedBiFunc<>(fnc);
+        this.fst = fst;
+        this.snd = snd;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return this.fst.hasNext() && this.snd.hasNext();
```
llorllale: @dgroup I think we would allow inconsistencies to pass if we don't check whether values of both `hasNext()` are equal

src/main/java/org/cactoos/iterator/Correlated.java
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    @Override
+    public X next() {
+        final X fvl = this.fst.next();
+        final X svl = this.snd.next();
+        if (this.fnc.apply(fvl, svl)) {
+            return fvl;
+        } else {
+            throw new NoSuchElementException(
```
llorllale: @dgroup I think `IllegalStateException` would be better, don't you think?

dgroup: From IllegalStateException docs
```java
/**
 * Signals that a method has been invoked at an illegal or
 * inappropriate time.  In other words, the Java environment or
 * Java application is not in an appropriate state for the requested
 * operation.
 *
 * @author  Jonni Kanerva
 * @since   JDK1.1
 */
```
thus, I'm not sure that this is a good idea. 
`NoSuchElementException` means that element which satisfies the required condition/function is absent. It's closer from a semantic perspective then ISE.

llorllale:
@rultor merge

llorllale:
@dgroup thanks!

0crat:
Job `gh:yegor256/cactoos#961` is not assigned, can't get performer

0crat:
The job [#961](https://github.com/yegor256/cactoos/issues/961) is now out of scope

Issue #1046, opened 2019-01-27
package-info.java:29-33: Classes from package {@link...


0pdd:
The puzzle `1006-0071ae30` from #1006 has to be resolved: 

https://github.com/yegor256/cactoos/blob/6b4a6b9a2b0eeb8b7ab7e9c0e7d1e57bfda91784/src/main/java/org/cactoos/time/package-info.java#L29-L33

The puzzle was created by @rultor on 27-Jan-19. 

  

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
This puzzle inherited milestone `1.0` from issue #1006.

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
The job [#1046](https://github.com/yegor256/cactoos/issues/1046) assigned to @atapin[/z](https://www.0crat.com/u/atapin), here is [why](https://www.0crat.com/footprint/C63314D6Z/68bf87a0-543e-4b53-a771-58a6d20b30ab); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

0crat:
[...]

PR #1123, opened 2019-04-29
(#1046) Move asText classes from time package to TextOf

Umbrah:
This is for #1046 
I moved cosntructors from *asText classes in time package to TextOf


0crat:
Job [#1123](https://github.com/yegor256/cactoos/issues/1123) is now in scope, role is `REV`

0crat:
This pull request [#1123](https://github.com/yegor256/cactoos/issues/1123) is assigned to @vzurauskas[/z](https://www.0crat.com/u/vzurauskas), here is [why](https://www.0crat.com/footprint/C63314D6Z/698e4d11-fd1f-4662-9604-748535fea803); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @llorllale[/z](https://www.0crat.com/u/llorllale) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be no monetary reward for this job

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -399,6 +403,210 @@ public TextOf(final LocalDate date, final DateTimeFormatter formatter) {
...
+
+    /**
+     * Formats the date using the format and locale using the system default
+     * zone.
+     * @param date The date to format.
+     * @param format The format to use.
+     * @param locale The locale to use.
+     */
+    public TextOf(final Date date, final String format,
+        final Locale locale) {
```
vzurauskas: Line wrapping is inconsistent with other classes, e.g. `FormattedText`. I think we should use the ["bracket should either start/end a line or be paired on the same line"](https://www.yegor256.com/2014/10/23/paired-brackets-notation.html) rule (as in `FormattedText`).

So in this case it would look like this:

```java
public TextOf(
    final Date date,
    final String format,
    final Locale locale
) {
    . . .
}
```

This comment applies to other constructors as well.

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -399,6 +403,210 @@ public TextOf(final LocalDate date, final DateTimeFormatter formatter) {
...
+     * Formats the date using the format and locale using the system default
+     * zone.
+     * @param date The date to format.
+     * @param formatter The formatter to use.
+     */
+    public TextOf(final Date date, final DateTimeFormatter formatter) {
+        super((Scalar<String>) () -> new TextOf(
+            ZonedDateTime.ofInstant(
+                date.toInstant(),
+                ZoneId.of("UTC")
```
vzurauskas: Javadoc says system default zone will be used, but this is UTC.

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -399,6 +403,210 @@ public TextOf(final LocalDate date, final DateTimeFormatter formatter) {
...
+    public TextOf(final LocalDateTime date, final String format,
+        final Locale locale) {
+        this(date, DateTimeFormatter.ofPattern(format, locale));
+    }
+
+    /**
+     * Formats the milliseconds using the ISO format.
+     * @param milliseconds Milliseconds to format.
+     */
+    public TextOf(final Long milliseconds) {
```
vzurauskas: It is very unintuitive that `new TextOf(1234L)` is a textual date. Any reasonable user would expect it to just be a "1234L" string. And if not, why date specifically? A Long number could represent anything: distance, money, temperature, etc. I think formatting milliseconds to ISO date is a responsibility for a new class, `FormattedEpoch` or something like that. And then the no-arg `DateAsText` constructor could be moved there, as now it's lost in the migration.

This comment applies to other constructors taking milliseconds as well.

Umbrah:
@vzurauskas I fixed comments according to utc zone and constructors formatting. As for constructors with long parameters, I added to-do to create a class like you described and use it in that ctors.

vzurauskas:
@Umbrah I think we should leave the constructors with `Long` arguments in the classes they were in originally rather than moving them back and forth with this task and then the new one somebody is going to implement. It might be a long time until they are in the right place, and they made more sense in those original classes. Once somebody moves them to the new class, they can delete the old ones.

Also, in the task description I would specify that the new class is not a `Date`, but a textual representation of a date, a type of `Text`.

Umbrah:
@vzurauskas I agree, I will move them back and just add to-do to original classes. 

> Also, in the task description I would specify that the new class is not a Date, but a textual representation of a date, a type of Text.

I think this new class should represent a Date, so we can work with it like new EpochDate(1000L).value() - returns Date; 
And use this class in constructor of TextOf.
Otherwise we will get the same situation which this PR is supposed to fix - there will be a class representing Text but separate from TextOf and in different package. Please let me know what do you think about it.

vzurauskas:
@Umbrah That makes sence, although there already is a class like that in the JDK: [Date](https://docs.oracle.com/javase/7/docs/api/java/util/Date.html#Date(long)). So maybe there’s no need for a new class after all. Perhaps you should just update the todo with these considerations and let the developer who takes that task to decide?

Umbrah:
@vzurauskas Agree, we can just use new TextOf(new Date()), no need in separate class. I deleted those logic with millis from TextOf. In this case I don't think we need a need a to-do, maybe let's see what ARC think?

vzurauskas:
@Umbrah Alright.

@llorllale Looks good to me. Note the discussion we had about the constructors with millisecond arguments, see if you agree with this.

Umbrah:
@llorllale ping

llorllale:
@Umbrah @vzurauskas I agree that `new TextOf(long)` should print out the number in textual form.

I think I see a mismatch between part of what was discussed (create todo to add new classes/leave old classes with long parameters behind) and what is actually implemented.

But that's ok - this PR is fine as it is.

llorllale:
@rultor merge

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @vzurauskas[/z](https://www.0crat.com/u/vzurauskas), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
Code review was too long (10 days), architects (@llorllale) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

0crat:
The job [#1123](https://github.com/yegor256/cactoos/issues/1123) is now out of scope

0crat:
Payment to `ARC` for a closed pull request, as in [§28](http://www.zerocracy.com/policy.html#28): +10 point(s) just awarded to @llorllale[/z](https://www.0crat.com/u/llorllale)

sereshqua:
@0crat quality good

0crat:
Order was finished, quality is "good": +20 point(s) just awarded to @vzurauskas[/z](https://www.0crat.com/u/vzurauskas)

0crat:
Quality review completed: +4 point(s) just awarded to @sereshqua[/z](https://www.0crat.com/u/sereshqua)

Issue #1143, opened 2019-06-11
OutputOf(File) fails on mkdirs on relative file without parent directories (in relative path)


cocsackie:
# Minimal code to reproduce:
`new OutputTo(new File("a.txt")).stream();`
## Unexpected behaviour
> Exception in thread "main" java.lang.NullPointerException
> 	at org.cactoos.io.OutputTo.lambda$new$0(OutputTo.java:67)
> 	at org.cactoos.io.OutputTo.stream(OutputTo.java:187)
> 	at test.Cactoos.main(Cactoos.java:19)
> 
## Expected behaviour
Should work in the same way as when:
- file is `new File("a.txt").getAbsoluteFile()`
- file is `new File("a.txt").getCanonicalFile()`
# Remarks:

Relative files work:

`new OutputTo(new File("test/a.txt")).stream();`

The problem is in(notice no null check on File.getParentFile()):

```java
if (mkdirs) {
	file.getParentFile().mkdirs();
}
```

When the path is relative and it's just a name of a file, File.getParentFile() returns null.
It seems that it could be changed to (since passing `new File("a.txt").getAbsoluteFile()` works):

```java
if (mkdirs) {
	file.getAbsoluteFile().getParentFile().mkdirs();
}
```
It seems that `File.getAbsoluteFile()` is preferred to `File.getCanonicalFile()` due to `File.getCanonicalFile()` throwing `IOException`.

I didn't test this on Linux/MacOS and this might be system dependent.

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
@cocsackie[/z](https://www.0crat.com/u/cocsackie) this project will fix the problem faster if you donate a few dollars to it; just [click here](https://www.0crat.com/contrib/C63314D6Z) and pay via Stripe, it's very fast, convenient and appreciated; thanks a lot!

llorllale:
@0crat in

0crat:
> @0crat in ([here](https://github.com/yegor256/cactoos/issues/1143#issuecomment-503350075))

@llorllale Job [#1143](https://github.com/yegor256/cactoos/issues/1143) is now in scope, role is `DEV`

0crat:
> @0crat in ([here](https://github.com/yegor256/cactoos/issues/1143#issuecomment-503350075))

[...]

PR #1155, opened 2019-07-01
(#1143) Using getAbsoluteFile() to fix error for single filename files

scristalli:
This is for #1143.

There isn't much to say, the issue description already analyzes the problem in detail and suggests a solution. I agree, and applied the simple fix.
I see no way to correctly unit test this, since it would require the creation of a file from a relative path consisting only of the filename, and of course we would want to put such file in a temporary directory, like we do for other tests. However, it seems that [there is no reliable way to accomplish this](https://stackoverflow.com/questions/840190/changing-the-current-working-directory-in-java).

### DIFF
--- a/src/main/java/org/cactoos/io/OutputTo.java
+++ b/src/main/java/org/cactoos/io/OutputTo.java
@@ -64,7 +64,7 @@ public final class OutputTo implements Output {
         this(
             () -> {
                 if (mkdirs) {
-                    file.getParentFile().mkdirs();
+                    file.getAbsoluteFile().getParentFile().mkdirs();
                 }
                 return new FileOutputStream(file);
             }


0crat:
@scristalli[/z](https://www.0crat.com/u/scristalli) this pull request is too small, just 2 lines changed (less than 10), there will be no formal code review, see [§53](http://www.zerocracy.com/policy.html#53) and [§28](http://www.zerocracy.com/policy.html#28); in the future, try to make sure your pull requests are not too small; @llorllale[/z](https://www.0crat.com/u/llorllale) please review this and merge or reject

llorllale:
@rultor merge

llorllale:
@scristalli here's your error: `bd4242b: subject length exceeds max [72]`.

Protip: install [this](https://github.com/llorllale/go-gitlint) tool and set up a git `commit-msg` hook as per [these](https://github.com/llorllale/go-gitlint#lint-your-commit-msg-when-committing) instructions

scristalli:
@llorllale thanks, I saw the Rultor message. Does it refuse commit messages longer than 72 characters? I think that's strange because I can see commits with longer messages, for instance commit 1c6f455037b6. Anyway, I force-pushed a shorter message, should be fine now.
Thanks for the tip, I will proceed and use the tool ASAP to avoid problems like this in the future.

llorllale:
@scristalli commit linting is configured in [.gitlint](https://github.com/yegor256/cactoos/blob/master/.gitlint)

scristalli:
@llorllale commit d7590f7a7cca17030 that I pushed passes the go-gitlint locally with `gitlint --subject-regex="^\(#[0-9]+\) [^\s].{13,50}" --subject-maxlen=72 --since=2019-06-25`, can we retry merging?

llorllale:
@rultor merge

0crat:
Job `gh:yegor256/cactoos#1155` is not assigned, can't get performer

0crat:
This job is not in scope

Issue #898, opened 2018-05-22
AndInThreadsTest.java:49-51: Remove the use of the static...


0pdd:
The puzzle `829-59f865f8` from #829 has to be resolved:

https://github.com/yegor256/cactoos/blob/995eb06e36912fe63f3ab22d1f992284084ee698/src/test/java/org/cactoos/scalar/AndInThreadsTest.java#L49-L51

The puzzle was created by Paulo Benety on 22-May-18. 

Estimate: 30 minutes,     

If you have any technical questions, don't ask me, submit new tickets instead. The task will be "done" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
Job [#898](https://github.com/yegor256/cactoos/issues/898) is now in scope, role is `DEV`

0crat:
The job [#898](https://github.com/yegor256/cactoos/issues/898) assigned to @vssekorin[/z](https://www.0crat.com/u/vssekorin), here is [why](https://www.0crat.com/footprint/C63314D6Z/1851206082); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be a monetary reward for this job

0crat:
The user @vssekorin[/z](https://www.0crat.com/u/vssekorin) resigned from [#898](https://github.com/yegor256/cactoos/issues/898), please stop working. Reason for job resignation: It is older than 10 days, see [§8](http://www.zerocracy.com/policy.html#8)

0crat:
Job `gh:yegor256/cactoos#898` is not assigned, can't get performer

0crat:
Resigned on delay, see [§8](http://www.zerocracy.com/policy.html#8): -30 point(s) just awarded to @vssekorin[/z](https://www.0crat.com/u/vssekorin)

0crat:
[...]

PR #1231, opened 2019-11-09
(#1230) ListIteratorOf can be mutable and immutable, depending on origin

fanifieiev:
This PR is for [1230](https://github.com/yegor256/cactoos/issues/1230)
The [ListIteratorOf](https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/list/ListIteratorOf.java) should support mutability, if the origin supports

### DIFF
--- a/src/test/java/org/cactoos/list/ListEnvelopeTest.java
+++ b/src/test/java/org/cactoos/list/ListEnvelopeTest.java
@@ -45,7 +45,7 @@ import org.llorllale.cactoos.matchers.Assertion;
  *  That's because this test should check the original behavior of ListEnvelope
  *  Now this test checks behavior of the Immutable decorator
  */
-@SuppressWarnings({ "PMD.TooManyMethods", "PMD.AvoidDuplicateLiterals" })
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.AvoidDuplicateLiterals"})
 public final class ListEnvelopeTest {
 
     @Test(expected = UnsupportedOperationException.class)
@@ -159,9 +159,9 @@ public final class ListEnvelopeTest {
     }
 
     @Test
-    public void getsPreviousIndex() {
+    public void mustReturnPreviousIndex() {
         new Assertion<>(
-            "List iterator returns incorrect previous index",
+            "List Iterator must return previous index",
             new ListIteratorOf<>(
                 new ListOf<>(1)
             ).previousIndex(),
@@ -170,9 +170,9 @@ public final class ListEnvelopeTest {
     }
 
     @Test
-    public void getsPrevious() {
+    public void mustReturnPreviousElement() {
         new Assertion<>(
-            "List iterator returns incorrect previous item",
+            "List Iterator must return previous element",
             new ListIteratorOf<>(
                 new ListOf<>(3, 7),
                 1
@@ -182,9 +182,9 @@ public final class ListEnvelopeTest {
     }
 
     @Test
-    public void getsNextIndex() {
+    public void mustReturnNextIndex() {
         new Assertion<>(
-            "List iterator returns incorrect next index",
+            "List iterator must return next index",
             new ListIteratorOf<>(
                 new ListOf<>(1)
             ).nextIndex(),
@@ -193,9 +193,9 @@ public final class ListEnvelopeTest {
     }
 
     @Test
-    public void getsNext() {
+    public void mustReturnNextElement() {
         new Assertion<>(
-            "List iterator returns incorrect next item",
+            "List iterator must return next item",
             new ListIteratorOf<>(
                 new ListOf<>(5, 11, 13),
                 1
--- a/src/test/java/org/cactoos/list/NoNullsTest.java
+++ b/src/test/java/org/cactoos/list/NoNullsTest.java
@@ -152,7 +152,6 @@ public final class NoNullsTest {
                 return 0;
             },
             new Throws<>(
-                "Iterator is read-only and doesn't allow adding items",
                 UnsupportedOperationException.class
             )
         ).affirm();
@@ -167,7 +166,6 @@ public final class NoNullsTest {
                 return 0;
             },
             new Throws<>(
-                "Iterator is read-only and doesn't allow removing items",
                 UnsupportedOperationException.class
             )
         ).affirm();
@@ -182,7 +180,6 @@ public final class NoNullsTest {
                 return 0;
             },
             new Throws<>(
-                "Iterator is read-only and doesn't allow rewriting items",
                 UnsupportedOperationException.class
             )
         ).affirm();
--- a/src/main/java/org/cactoos/list/ListIteratorOf.java
+++ b/src/main/java/org/cactoos/list/ListIteratorOf.java
@@ -30,15 +30,13 @@ import org.cactoos.scalar.Sticky;
 import org.cactoos.scalar.Unchecked;
 
 /**
- * Iterator of the list that doesn't allow mutations.
+ * List Iterator of the list that allows mutations if the original
+ *  list iterator does support.
  *
  * <p>There is no thread-safety guarantee.
  *
  * @param <T> Items type
  * @since 0.35
- * @todo #1219:30min {@link ListIteratorOf} does not have to be immutable.
- *  Make it possible to use mutable operations like remove/set/add in this
- *  {@link ListIteratorOf} and change the class javadoc as well.
  */
 public final class ListIteratorOf<T> implements ListIterator<T> {
 
@@ -112,22 +110,16 @@ public final class ListIteratorOf<T> implements ListIterator<T> {
 
     @Override
     public void remove() {
-        throw new UnsupportedOperationException(
-            "Iterator is read-only and doesn't allow removing items"
-        );
+        this.origin.value().remove();
     }
 
     @Override
     public void set(final T item) {
-        throw new UnsupportedOperationException(
-            "Iterator is read-only and doesn't allow rewriting items"
-        );
+        this.origin.value().set(item);
     }
 
     @Override
     public void add(final T item) {
-        throw new UnsupportedOperationException(
-            "Iterator is read-only and doesn't allow adding items"
-        );
+        this.origin.value().add(item);
     }
 }
--- a/src/main/java/org/cactoos/list/package-info.java
+++ b/src/main/java/org/cactoos/list/package-info.java
@@ -26,5 +26,11 @@
  * Lists, tests.
  *
  * @since 0.14
+ * @todo #1230:30min The following list implmenetation classes
+ *  {@link org.cactoos.list.Joined}, {@link org.cactoos.list.Mapped},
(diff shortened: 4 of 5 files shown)


0crat:
Job [#1231](https://github.com/yegor256/cactoos/issues/1231) is now in scope, role is `REV`

0crat:
This pull request [#1231](https://github.com/yegor256/cactoos/issues/1231) is assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C63314D6Z/f31c631f-6d4a-4b7e-ab97-f9f52a3230f7); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @paulodamaso[/z](https://www.0crat.com/u/paulodamaso) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be no monetary reward for this job

src/test/java/org/cactoos/list/ListEnvelopeTest.java
```diff
@@ -159,9 +159,9 @@ public void returnsSubListWithUnsupportedAdd() {
     }
 
     @Test
-    public void getsPreviousIndex() {
+    public void mustReturnPreviousIndex() {
         new Assertion<>(
-            "List iterator returns incorrect previous index",
+            "List Iterator must return previous index",
```
victornoel: @fanifieiev why is `ListIteratorOf` tested in `ListEnvelopeTest`? I think those tests should be in their own file (and removed from here).

src/test/java/org/cactoos/list/NoNullsTest.java
```diff
@@ -167,7 +166,6 @@ public void removeThrowsErrorForImmutableListIterator() {
                 return 0;
             },
             new Throws<>(
-                "Iterator is read-only and doesn't allow removing items",
```
victornoel: @fanifieiev why remove those parts of the tests?

src/test/java/org/cactoos/list/ListEnvelopeTest.java
fanifieiev: @victornoel I agree with your concern, I just left them where they were previously.

src/test/java/org/cactoos/list/NoNullsTest.java
fanifieiev: @victornoel If you look at the test and the following stacktrace, you can see that the exception is thrown by the collection returned by `Collections.unmodifiableList`, that is why it throws exception with no message. 
image

fanifieiev:
> @fanifieiev @paulodamaso I question the very existence of `ListOteratorOf`, it does nothing except wrapping a `ListIterator`: either we should remove it or we can rename it `ListIteratorEnvelope` and make it `abstract`. The later could make sense I believe. Please decide @paulodamaso.

@victornoel I would keep it as an abstract class ONLY if we really need it, otherwise would remove.

src/test/java/org/cactoos/list/ListEnvelopeTest.java
victornoel: @fanifieiev so please move them in their own file to properly identify what is being tested

src/test/java/org/cactoos/list/NoNullsTest.java
victornoel: @fanifieiev I better understand, thank you, but why is `ListIteratorNoNulls` based on an immutable version of `ListIterator`? I think that it is coherent we the changes we are making that `ListIteratorNoNulls` is to be only concerned with nulls, not with mutability. Please fix it or add a todo to continue the work.

fanifieiev: @victornoel I did not catch your concern. The  [ListIteratorOf](https://github.com/yegor256/cactoos/pull/1231/commits/52758f3b87a55249bdfdff4e8ae76f7847bb7285#diff-938af5d32ea1b9126e592148160ddbe9) I changed  is not based on the immutable version of `ListIterator`.
Could you provide details regarding you concern?

victornoel: I'm talking about `ListIteratorNoNulls`: those tests on which we are currently commenting seem to validate that it is immutable but I don't think it is in the spirit of the changes that are being done to the `ListIterator` type hierarchy in this PR: either change `ListIteratorNoNulls` to be coherent with the changes you are currently making, or add a todo in order for someone to do it.

fanifieiev: @victornoel The test we are currently commenting on is [tests](https://github.com/yegor256/cactoos/pull/1231/files/52758f3b87a55249bdfdff4e8ae76f7847bb7285#diff-f3ddcee3c2a90c165c537d1734be8fdc).
If yes, then those tests are not using 'ListIteratorNoNulls', they use 'NoNulls' list and the reason why this behavior is immutable, is that 'ListOf' uses 'Collections.unmodifiableList'.
So, there is nothing related to 'ListIteratorNoNulls'.

victornoel: @fanifieiev I understand, ok. But in any case, this is related to the work you are doing right now (that `ListIteratorOf` should not enforce immutability) and I think it would be relevant to continue this work by adding a todo. So in this case that `list.NoNulls` nor `ListOf`, nor `ListIterableNoNulls` should not enforce immutability. It's not because we are working on a small tasks that we shouldn't think about the bigger picture, this is the difference between chaos and iterative/incremental development.

victornoel: @fanifieiev as a side comment: as the DEV working on this, you are the best person to evaluate what are the next changes to continue implemen the vision that the current task you are doing is promoting.

fanifieiev: @victornoel Now your concern is clear to me.
What if I make changes to all mentioned classes(list.NoNulls nor ListOf, nor ListIterableNoNulls should not enforce immutability) in scope of this PR ?
It will be faster, than creating a todo and waiting  for someone fix/review.

victornoel: @fanifieiev it's good for me, your choice :)

fanifieiev: @victornoel @paulodamaso A lot of list implementations in the `org.cactoos.list` package construct `Collections.unmodifiableList` which makes the lists immutable.
In my opinion we do not have to enforce them be immutable, but rather mutable and each of them bears its own specific responsibility. 
If someone wants to construct, for example, an immutable 'org.cactoos.list.Sorted' list, he/she should decorate it with `org.cactoos.list.Immutable`: `new Immutable<>(new Sorted(10, 2, 3, 5))`.
So, using this approach, we can provide API, that can be used in both mutable/immutable cases.
I am not 100% sure this idea is prefect, as the `Collections.unmodifableList` was added by @yegor256 in the above mentioned lists.
Please share you thoughts.

victornoel: @fanifieiev to me you are rephrasing what I already said and what was decided by ARC with the accord of @yegor256 previously in #898. You just need to create the todos et and let's get on with it.

fanifieiev: @victornoel I can fix that in scope of current PR if this is acceptable.

victornoel: @fanifieiev I think we spent enough time on this PR, please add todos and let ARC validate them by merging.

victornoel: @fanifieiev one todo is enough, no need to multiplicate them btw :)

fanifieiev: @victornoel There is one thing that concerns me.
Let's say we will make those mentioned classes mutable. Let's take `Sorted` as an example:
`List<Integer> sorted = new Sorted<>(2, 1, 0, 5);` //sorted list is now sorted
Let's add one more element:
`sorted.add(6);`
//sorted list is not sorted anymore
You got the idea of my concern?

victornoel: @fanifieiev that's why you should leave a todo and let the next job performer think about this problem. Maybe we will conclude that `Sorted` can't be mutable, but that's out of scope of the present issue. Personally I believe the real problem is that it's not relevant to decorate collection to sort them, but that's only my opinion ;) Also, nothing prevent to implement Sorted using a Java ordered list and implement insertion in a sorted manner.

fanifieiev: @victornoel Ok, I have just pushed one more commit with 'todo'. Please have a look.

victornoel:
@paulodamaso it's good on my side, please take a look and as the [above](https://github.com/yegor256/cactoos/pull/1231#discussion_r344458428) discussion as well and to my previous [comment](https://github.com/yegor256/cactoos/pull/1231#pullrequestreview-314573667) too.

paulodamaso:
@victornoel @fanifieiev It's good, thank you.

paulodamaso:
@rultor merge

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @victornoel[/z](https://www.0crat.com/u/victornoel), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
The job [#1231](https://github.com/yegor256/cactoos/issues/1231) is now out of scope

0crat:
Payment to `ARC` for a closed pull request, as in [§28](http://www.zerocracy.com/policy.html#28): +10 point(s) just awarded to @paulodamaso[/z](https://www.0crat.com/u/paulodamaso)

sereshqua:
@victornoel please make sure you would try to find at least 3 issues during next CR, thanks

victornoel:
@sereshqua well actually I did if you take the global review comment into account :)

sereshqua:
@victornoel good point, Policy does not give clear instruction and its treatment can be subjective;

from one point there are 3 comments: 1 global and 2 exact (pointing the code out),
from another point: should the global comment be taken into account at all or not?

victornoel:
@sereshqua hehe, do as you see fit, it's not very important in the end, typically there is no ambiguities :)

sereshqua:
@0crat quality acceptable

0crat:
Order was finished, quality is "acceptable": +15 point(s) just awarded to @victornoel[/z](https://www.0crat.com/u/victornoel)

0crat:
Quality review completed: +4 point(s) just awarded to @sereshqua[/z](https://www.0crat.com/u/sereshqua)

PR #1311, opened 2020-02-23
Remove some collection decorators

victornoel:
This is for #1296, #1297 and #1303.

I did all at once because the changes are very limited to the removed classes.

Issue #1304, opened 2020-02-11
TailOf.java:35-38: Remove this class and replace it...


0pdd:
The puzzle `1242-1b85f1d7` from #1242 has to be resolved: 

https://github.com/yegor256/cactoos/blob/2a62719ec67ee153ff99f2cbb9aacb9429efbea4/src/main/java/org/cactoos/collection/TailOf.java#L35-L38

The puzzle was created by Victor Noël on 09-Feb-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0crat:
The job [#1304](https://github.com/yegor256/cactoos/issues/1304) assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C63314D6Z/dc826687-4ab1-4ae5-83dd-a598ae255059); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

victornoel:
@0crat wait for #1315 to be merged

0crat:
> @0crat wait for #1315 to be merged ([here](https://github.com/yegor256/cactoos/issues/1304#issuecomment-592964956))

@victornoel The impediment for [#1304](https://github.com/yegor256/cactoos/issues/1304) was registered successfully by @victornoel[/z](https://www.0crat.com/u/victornoel)

0pdd:
The puzzle `1242-1b85f1d7` has disappeared from the source code, that's why I closed this issue.

victornoel:
@paulodamaso can you do something to trigger QA here?

victornoel:
@0crat status

[...]

Issue #1303, opened 2020-02-11
Skipped.java:35-39: Remove this class and replace it...


0pdd:
The puzzle `1242-15741ade` from #1242 has to be resolved: 

https://github.com/yegor256/cactoos/blob/2a62719ec67ee153ff99f2cbb9aacb9429efbea4/src/main/java/org/cactoos/collection/Skipped.java#L35-L39

The puzzle was created by Victor Noël on 09-Feb-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0crat:
The job [#1303](https://github.com/yegor256/cactoos/issues/1303) assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C63314D6Z/1849ef81-94e3-47f8-91f8-55b28f5c6019); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

victornoel:
@0crat wait for #1311 to be merged

0crat:
> @0crat wait for #1311 to be merged ([here](https://github.com/yegor256/cactoos/issues/1303#issuecomment-590050418))

@victornoel The impediment for [#1303](https://github.com/yegor256/cactoos/issues/1303) was registered successfully by @victornoel[/z](https://www.0crat.com/u/victornoel)

0pdd:
The puzzle `1242-15741ade` has disappeared from the source code, that's why I closed this issue.

0crat:
[...]

Issue #1299, opened 2020-02-11
Filtered.java:36-40: Remove this class and replace it...


0pdd:
The puzzle `1242-d08d3037` from #1242 has to be resolved: 

https://github.com/yegor256/cactoos/blob/2a62719ec67ee153ff99f2cbb9aacb9429efbea4/src/main/java/org/cactoos/collection/Filtered.java#L36-L40

The puzzle was created by Victor Noël on 09-Feb-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0crat:
The job [#1299](https://github.com/yegor256/cactoos/issues/1299) assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C63314D6Z/b030a5cb-4d1b-4d5d-9604-2504d5aae109); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

victornoel:
@0crat wait for #1315 to be merged

0crat:
> @0crat wait for #1315 to be merged ([here](https://github.com/yegor256/cactoos/issues/1299#issuecomment-592964951))

@victornoel The impediment for [#1299](https://github.com/yegor256/cactoos/issues/1299) was registered successfully by @victornoel[/z](https://www.0crat.com/u/victornoel)

0pdd:
The puzzle `1242-d08d3037` has disappeared from the source code, that's why I closed this issue.

0crat:
[...]

PR #1315, opened 2020-02-29
Remove some collection decorators

victornoel:
This is for #1299 and #1304. 

Changes:
- I moved when possible the tests from collection to iterable tests
- I added some of the test that were removed in #1311 for #1303 to iterable's `Skipped` since it seemed a good idea


0crat:
This pull request [#1315](https://github.com/yegor256/cactoos/issues/1315) is assigned to @fanifieiev[/z](https://www.0crat.com/u/fanifieiev), here is [why](https://www.0crat.com/footprint/C63314D6Z/d6bb12e0-4f33-489c-a436-ddbc070bf581); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @paulodamaso[/z](https://www.0crat.com/u/paulodamaso) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be no monetary reward for this job

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
     @Test
-    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
     public void skipIterable() {
-        MatcherAssert.assertThat(
-            "Can't skip elements in iterable",
+        new Assertion<>(
+            "Must skip elements in iterable",
+            new Skipped<>(
+                2,
+                new IterableOf<>("one", "two", "three", "four")
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+            ),
+            new IsEqual<>(
+                new IterableOf<>(
+                    "three",
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+        ).affirm();
+    }
+
+    @Test
+    public void skipArray() {
+        new Assertion<>(
+            "Must skip elements in array",
+            new Skipped<>(
+                2,
+                "one", "two", "three", "four"
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+            ),
+            new IsEqual<>(
+                new IterableOf<>(
+                    "three",
+                    "four"
+                )
+            )
+        ).affirm();
+    }
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+
+    @Test
+    public void skipCollection() {
+        new Assertion<>(
+            "Must skip elements in collection",
+            new Skipped<>(
+                2,
+                new ListOf<>("one", "two", "three", "four")
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+            ),
+            new IsEqual<>(
+                new IterableOf<>(
+                    "three",
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+        ).affirm();
+    }
+
+    @Test
+    public void skippedAllElements() {
+        new Assertion<>(
+            "Must skip all elements",
             new Skipped<>(
                 2,
+                "one", "two"
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+        ).affirm();
+    }
+
+    @Test
+    public void skippedMoreThanExists() {
+        new Assertion<>(
+            "Can't skip more than exists",
+            new Skipped<>(
+                Integer.MAX_VALUE,
+                "one", "two"
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
+    }
+
+    @Test
+    @SuppressWarnings("PMD.AvoidDuplicateLiterals")
+    public void skippedNegativeSize() {
+        new Assertion<>(
+            "Must process negative skipped size",
+            new Skipped<>(
+                -1,
                 "one", "two", "three", "four"
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
             ),
-            Matchers.contains(
-                "three",
-                "four"
+            new IsEqual<>(
+                new IterableOf<>(
+                    "one",
```
fanifieiev: @victornoel Please extract these values into constants.

src/test/java/org/cactoos/iterable/SkippedTest.java
```diff
@@ -23,31 +23,113 @@
...
 import org.junit.Test;
+import org.llorllale.cactoos.matchers.Assertion;
 
 /**
  * Test Case for {@link Skipped}.
  *
  * @since 0.34
  * @checkstyle JavadocMethodCheck (500 lines)
  */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
```
fanifieiev: @victornoel Please remove this one after extracting values into constants

victornoel:
@fanifieiev using constants in tests is counter-productive, each test should be autonomous, so using constants:
- prevent that
- makes the code less readable
- does not provide anything except removing one suppress warning that is clearly not made for tests code

See https://www.yegor256.com/2016/05/03/test-methods-must-share-nothing.html for more details

victornoel:
@paulodamaso please consider the comment above too, it is not the first time that I see such requests from REVs and I really believe it's going against the well being of the project to blindly try to follow all the warnings of PMD. Let's not mixup the spirit of the law with the letter of the law please…

fanifieiev:
> @paulodamaso please consider the comment above too, it is not the first time that I see such requests from REVs and I really believe it's going against the well being of the project to blindly try to follow all the warnings of PMD. Let's not mixup the spirit of the law with the letter of the law please…

@victornoel @paulodamaso Well, we have to decide here and follow the same way everytime, because I recently had to fix similar issues raised by @paulodamaso .

victornoel:
@paulodamaso ping?

paulodamaso:
@victornoel I am in favor of constant usages for repetitive data, that's why:
- most of our test cases already share test data. Let me explain: we use the same values on almost all of our test cases (e.g. `0, 1, 2, "one", "two", "three"`) and these values are repeated. Why not put them into constants if they are present into most of our test cases if they are the same? Or else, you could alternatively do not use the same set of data for each test, avoiding the duplication.
- we can use context-sensitive naming for better test understanding: we have lots of criptic statuses and return codes which are not too obvious to the programmer (take HTTP statuses for example). It is much better to check a status against a constant with meaning (for example, `HTTP_OK`) instead some magic number (`200`). The first is much more readable than the latter.
- we have some repeatable data that can fit into constants, avoiding mistakes: `"text/plain"`, `"multipart/formdata"` and so on are examples of repeatable values that can be shared across tests and should be turned into constants.

Makes sense?

fanifieiev:
@victornoel ping

victornoel:
@paulodamaso I'm a bit bothered because I don't agree at all with what you presented, I think it doesn't make sense and it is useless work just for the appearance of quality while it will actually reduce the quality of the code.

I propose to either not do it, or add a `@todo` for removing the suppress warning and avoid tests inputs duplication by using different literals. I don't think it is a good idea to use literal constant for this. Here is why:

> most of our test cases already share test data. Let me explain: we use the same values on almost all of our test cases (e.g. 0, 1, 2, "one", "two", "three") and these values are repeated. Why not put them into constants if they are present into most of our test cases if they are the same? Or else, you could alternatively do not use the same set of data for each test, avoiding the duplication

Each tests rely on similar constants, but in terms of domain, they are just similar "by chance", not by design. This would be a mistake to intertwine all the tests because of that. And it will take much useles energy to avoid using the same terms (but it can be an acceptable solution though).

> we can use context-sensitive naming for better test understanding: we have lots of criptic statuses and return codes which are not too obvious to the programmer (take HTTP statuses for example). It is much better to check a status against a constant with meaning (for example, HTTP_OK) instead some magic number (200). The first is much more readable than the latter.
> we have some repeatable data that can fit into constants, avoiding mistakes: "text/plain", "multipart/formdata" and so on are examples of repeatable values that can be shared across tests and should be turned into constants.

I don't think this has anything to do with the constant used in the tests, in that case the duplication is duplication of the same domain concept, so it makes sense to factor it. FYI, even using literal constant for HTTP codes is not something that @yegor256 would recommend, see https://www.yegor256.com/2015/07/06/public-static-literals.html.

paulodamaso:
> @paulodamaso I'm a bit bothered because I don't agree at all with what you presented, I think it doesn't make sense and it is useless work just for the appearance of quality while it will actually reduce the quality of the code.
> 
> I propose to either not do it, or add a `@todo` for removing the suppress warning and avoid tests inputs duplication by using different literals. I don't think it is a good idea to use literal constant for this. Here is why:
> 
> > most of our test cases already share test data. Let me explain: we use the same values on almost all of our test cases (e.g. 0, 1, 2, "one", "two", "three") and these values are repeated. Why not put them into constants if they are present into most of our test cases if they are the same? Or else, you could alternatively do not use the same set of data for each test, avoiding the duplication
> 
> Each tests rely on similar constants, but in terms of domain, they are just similar "by chance", not by design. This would be a mistake to intertwine all the tests because of that. And it will take much useles energy to avoid using the same terms (but it can be an acceptable solution though).

@victornoel I think that you understand perfectly what I meant: if these values are the same "by chance", why  not make them explicitly different then? I agree with you when you say that it would be a mistake to tie all tests, but my point is, using constants or not, when repeating the same we are already coupling those tests around these values. In fact, it just becomes clearer when we get to the conclusion that it is not a good idea to use a constant for them. So, pleae leave a puzzle so we can fix it later.
> 
> > we can use context-sensitive naming for better test understanding: we have lots of criptic statuses and return codes which are not too obvious to the programmer (take HTTP statuses for example). It is much better to check a status against a constant with meaning (for example, HTTP_OK) instead some magic number (200). The first is much more readable than the latter.
> > we have some repeatable data that can fit into constants, avoiding mistakes: "text/plain", "multipart/formdata" and so on are examples of repeatable values that can be shared across tests and should be turned into constants.
> 
> I don't think this has anything to do with the constant used in the tests, in that case the duplication is duplication of the same domain concept, so it makes sense to factor it. FYI, even using literal constant for HTTP codes is not something that @yegor256 would recommend, see https://www.yegor256.com/2015/07/06/public-static-literals.html.
@victornoel Yes, the optimal solution would be create classes for each of these members in cases which the values are the same in the domain context.

victornoel:
@paulodamaso So if I understand well: it makes sense to have different literals in tests to avoid them being tied (by literal or by using constant, whatever) and have less strong tests?
This makes sense to me, I will add a todo in that direction :)

victornoel:
@fanifieiev @paulodamaso thanks, it is done

paulodamaso:
@rultor merge

paulodamaso:
> @paulodamaso So if I understand well: it makes sense to have different literals in tests to avoid them being tied (by literal or by using constant, whatever) and have less strong tests?
> This makes sense to me, I will add a todo in that direction :)

@victornoel what do you mean by "less strong tests"?

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @fanifieiev[/z](https://www.0crat.com/u/fanifieiev), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
Code review was too long (16 days), architects (@paulodamaso) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

sereshqua:
@0crat quality good

victornoel:
> @victornoel what do you mean by "less strong tests"?

@paulodamaso I meant what I think you were meaning: that if we use different literals for each test, that would make the test more robust.

Issue #1283, opened 2020-02-05
Zipping directory is broken


piotrkot:
Let's zip some files.
```
        final Path folder = new TempFolder("abc").value();
        new LengthOf(
            new TeeInput(
                new InputOf("A"),
                new OutputTo(
                    new File(folder.toFile(), "A.txt")
                )
            )
        ).value();
        new LengthOf(
            new TeeInput(
                new InputOf("B"),
                new OutputTo(
                    new File(folder.toFile(), "B.txt")
                )
            )
        ).value();
        final Path path = new TempFile("abc-compressed-", ".zip").value();
        new LengthOf(
            new TeeInput(
                new InputOf(
                    new Zip(new Directory(folder.toFile())).stream()
                ),
                new OutputTo(path)
            )
        ).value();
```
Files `A.txt` and `B.txt` have expected content and are located in the `/tmp/abc` directory.
Viewing contents with `unzip -l abc-compressed-8773065996033619463.zip` shows
```
Archive:  abc-compressed-8773065996033619463.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2020-02-05 20:52   /tmp/abc
        1  2020-02-05 20:52   /tmp/abc/A.txt
        1  2020-02-05 20:52   /tmp/abc/B.txt
---------                     -------
        2                     3 files
```
which is not expected. I didn't not expect to see `/tmp` directory in the zip archive.
When unzipping the contents with `unzip abc-compressed-8773065996033619463.zip -d unzipped/.` I get errors and warnings:
```
Archive:  abc-compressed-8773065996033619463.zip
warning:  stripped absolute path spec from /tmp/abc
  inflating: unzipped/./tmp/abc         
warning:  stripped absolute path spec from /tmp/abc/A.txt
checkdir error:  unzipped/./tmp/abc exists but is not directory
                 unable to process tmp/abc/A.txt.
warning:  stripped absolute path spec from /tmp/abc/B.txt
checkdir error:  unzipped/./tmp/abc exists but is not directory
                 unable to process tmp/abc/B.txt.
```
Also viewing contents with archive manager (file-roller) shows zipped additional `/tmp` directory and inside additional 'abc' empty file.

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

0crat:
[...]

PR #1321, opened 2020-03-11
(#1283) Fixed Zip input implementation

fanifieiev:
This PR is for issue #1283

### DIFF
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -30,13 +30,10 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.nio.file.Path;
-import java.util.Iterator;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import org.cactoos.Input;
 import org.cactoos.func.ForEach;
-import org.cactoos.iterator.Skipped;
-import org.cactoos.text.Joined;
 
 /**
  * Zip files and directory.
@@ -45,6 +42,9 @@ import org.cactoos.text.Joined;
  *
  * @since 0.29
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ * @todo #898:30min Currently {@link Zip} does not support zipping of empty
+ *  sub folder. Please change the implementation so that empty folders are
+ *  zipped as well.
  */
 @SuppressWarnings("PMD.AvoidInstantiatingObjectsInLoops")
 public final class Zip implements Input {
@@ -76,18 +76,6 @@ public final class Zip implements Input {
                         zip.putNextEntry(new ZipEntry(relative));
                         try (FileInputStream fis = new FileInputStream(file)) {
                             zip.write(new BytesOf(new InputOf(fis)).asBytes());
-                        }
-                    } else {
-                        final Iterator<Path> paths = new Skipped<>(
-                            1,
-                            new Directory(file).iterator()
-                        );
-                        if (!paths.hasNext()) {
-                            zip.putNextEntry(
-                                new ZipEntry(
-                                    new Joined("/", relative, "").asString()
-                                )
-                            );
                         }
                     }
                     zip.closeEntry();
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -30,10 +30,13 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.nio.file.Path;
+import java.util.Iterator;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import org.cactoos.Input;
 import org.cactoos.func.ForEach;
+import org.cactoos.iterator.Skipped;
+import org.cactoos.text.Joined;
 
 /**
  * Zip files and directory.
@@ -42,9 +45,6 @@ import org.cactoos.func.ForEach;
  *
  * @since 0.29
  * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
- * @todo #898:30min Currently {@link Zip} does not support zipping of empty
- *  sub folder. Please change the implementation so that empty folders are
- *  zipped as well.
  */
 @SuppressWarnings("PMD.AvoidInstantiatingObjectsInLoops")
 public final class Zip implements Input {
@@ -77,8 +77,20 @@ public final class Zip implements Input {
                         try (FileInputStream fis = new FileInputStream(file)) {
                             zip.write(new BytesOf(new InputOf(fis)).asBytes());
                         }
-                        zip.closeEntry();
+                    } else {
+                        final Iterator<Path> paths = new Skipped<>(
+                            1,
+                            new Directory(file).iterator()
+                        );
+                        if (!paths.hasNext()) {
+                            zip.putNextEntry(
+                                new ZipEntry(
+                                    new Joined("/", relative, "").asString()
+                                )
+                            );
+                        }
                     }
+                    zip.closeEntry();
                 }
             ).exec(this.directory);
         }
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -30,10 +30,12 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.InputStream;
 import java.nio.file.Path;
+import java.util.Iterator;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import org.cactoos.Input;
 import org.cactoos.func.ForEach;
+import org.cactoos.iterator.Skipped;
 import org.cactoos.scalar.ItemAt;
 import org.cactoos.text.Joined;
 
@@ -69,18 +71,24 @@ public final class Zip implements Input {
         try (ZipOutputStream zip = new ZipOutputStream(out)) {
             new ForEach<Path>(
                 path -> {
-                    final File file = path.toFile();
                     final String relative = parent.relativize(path).toString();
+                    final File file = path.toFile();
                     if (file.isFile()) {
                         zip.putNextEntry(new ZipEntry(relative));
                         try (FileInputStream fis = new FileInputStream(file)) {
                             zip.write(new BytesOf(new InputOf(fis)).asBytes());
                         }
-                    } else if (!new Directory(file).iterator().hasNext()) {
-                        final Joined join = new Joined(
-                            File.separator, relative, ""
+                    } else {
+                        final Iterator<Path> paths = new Skipped<>(
+                            1,
+                            new Directory(file).iterator()
                         );
-                        zip.putNextEntry(new ZipEntry(join.toString()));
+                        if (!paths.hasNext()) {
+                            final org.cactoos.text.Joined join = new Joined(
+                                "/", relative, ""
+                            );
+                            zip.putNextEntry(new ZipEntry(join.toString()));
+                        }
                     }
                     zip.closeEntry();
                 }
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -83,10 +83,11 @@ public final class Zip implements Input {
                             new Directory(file).iterator()
                         );
                         if (!paths.hasNext()) {
-                            final org.cactoos.text.Joined join = new Joined(
-                                "/", relative, ""
+                            zip.putNextEntry(
+                                new ZipEntry(
+                                    new Joined("/", relative, "").asString()
+                                )
                             );
-                            zip.putNextEntry(new ZipEntry(join.toString()));
                         }
                     }
                     zip.closeEntry();
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -77,6 +77,7 @@ public final class Zip implements Input {
                         try (FileInputStream fis = new FileInputStream(file)) {
                             zip.write(new BytesOf(new InputOf(fis)).asBytes());
                         }
+                        zip.closeEntry();
                     } else {
                         final Iterator<Path> paths = new Skipped<>(
                             1,
@@ -88,9 +89,10 @@ public final class Zip implements Input {
                                     new Joined("/", relative, "").asString()
                                 )
                             );
+                            zip.closeEntry();
                         }
                     }
-                    zip.closeEntry();
+
                 }
             ).exec(this.directory);
         }
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -77,8 +77,8 @@ public final class Zip implements Input {
                         try (FileInputStream fis = new FileInputStream(file)) {
                             zip.write(new BytesOf(new InputOf(fis)).asBytes());
                         }
+                        zip.closeEntry();
                     }
-                    zip.closeEntry();
                 }
             ).exec(this.directory);
         }
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -33,6 +33,9 @@ import java.nio.file.Path;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 import org.cactoos.Input;
+import org.cactoos.func.ForEach;
+import org.cactoos.scalar.ItemAt;
+import org.cactoos.text.Joined;
 
 /**
  * Zip files and directory.
@@ -40,43 +43,48 @@ import org.cactoos.Input;
  * <br>There is no thread-safety guarantee.
  *
  * @since 0.29
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
 @SuppressWarnings("PMD.AvoidInstantiatingObjectsInLoops")
 public final class Zip implements Input {
 
     /**
-     * Origin directory.
+     * Origin list of files.
      */
-    private final Directory origin;
+    private final Directory directory;
 
     /**
      * Ctor.
      *
-     * @param origin Origin directory.
+     * @param dir Origin directory.
      */
-    public Zip(final Directory origin) {
-        this.origin = origin;
+    public Zip(final Directory dir) {
+        this.directory = dir;
     }
 
     @Override
     public InputStream stream() throws Exception {
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        final Path parent = new ItemAt<>(0, this.directory).value().getParent();
         try (ZipOutputStream zip = new ZipOutputStream(out)) {
-            for (final Path path : this.origin) {
-                final File file = path.toFile();
-                final ZipEntry entry = new ZipEntry(
-                    file.getPath()
-                );
-                zip.putNextEntry(entry);
-                if (file.isFile()) {
-                    try (
-                        FileInputStream input = new FileInputStream(file)
-                    ) {
-                        zip.write(new BytesOf(new InputOf(input)).asBytes());
+            new ForEach<Path>(
+                path -> {
+                    final File file = path.toFile();
+                    final String relative = parent.relativize(path).toString();
+                    if (file.isFile()) {
+                        zip.putNextEntry(new ZipEntry(relative));
+                        try (FileInputStream fis = new FileInputStream(file)) {
+                            zip.write(new BytesOf(new InputOf(fis)).asBytes());
+                        }
+                    } else if (!new Directory(file).iterator().hasNext()) {
+                        final Joined join = new Joined(
+                            File.separator, relative, ""
+                        );
+                        zip.putNextEntry(new ZipEntry(join.toString()));
                     }
+                    zip.closeEntry();
                 }
-                zip.closeEntry();
-            }
+            ).exec(this.directory);
         }
         return new ByteArrayInputStream(out.toByteArray());
     }
--- a/src/main/java/org/cactoos/io/Zip.java
+++ b/src/main/java/org/cactoos/io/Zip.java
@@ -36,7 +36,6 @@ import java.util.zip.ZipOutputStream;
 import org.cactoos.Input;
 import org.cactoos.func.ForEach;
 import org.cactoos.iterator.Skipped;
-import org.cactoos.scalar.ItemAt;
 import org.cactoos.text.Joined;
 
 /**
@@ -67,11 +66,11 @@ public final class Zip implements Input {
     @Override
     public InputStream stream() throws Exception {
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
-        final Path parent = new ItemAt<>(0, this.directory).value().getParent();
         try (ZipOutputStream zip = new ZipOutputStream(out)) {
             new ForEach<Path>(
                 path -> {
-                    final String relative = parent.relativize(path).toString();
+                    final String relative = this.directory.path().getParent()
+                        .relativize(path).toString();
                     final File file = path.toFile();
                     if (file.isFile()) {
                         zip.putNextEntry(new ZipEntry(relative));
--- a/src/test/java/org/cactoos/io/ZipTest.java
+++ b/src/test/java/org/cactoos/io/ZipTest.java
@@ -27,6 +27,7 @@ package org.cactoos.io;
 import java.io.File;
 import java.io.OutputStream;
 import java.util.stream.Collectors;
+import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import org.cactoos.iterable.IterableOf;
 import org.cactoos.scalar.Sticky;
@@ -55,24 +56,49 @@ public final class ZipTest {
     public final TemporaryFolder temporal = new TemporaryFolder();
 
     @Test
+    public void mustZipEmptyDirectory() throws Exception {
+        final File folder = this.temporal.newFolder("empty");
+        new Assertion<>(
+            "Must zip empty directory",
+            new Sticky<>(
+                () -> {
+                    try (OutputStream out = new OutputStreamTo(
+                        this.temporal.newFile("empty.zip")
+                    )) {
+                        out.write(
+                            new BytesOf(
+                                new Zip(new Directory(folder))
+                            ).asBytes()
+                        );
+                    }
+                    try (ZipFile zipped = new ZipFile(
+                        new File(folder.getParentFile(), "empty.zip")
+                    )
+                    ) {
+                        return zipped.stream().map(
+                            ZipEntry::toString
+                        ).findFirst().get();
+                    }
+                }
+            ),
+            new ScalarHasValue<>(
+                "empty/"
+            )
+        ).affirm();
+    }
+
+    @Test
     public void mustZipDirectory() throws Exception {
         final String zipname = "abc.zip";
         final File folder = this.temporal.newFolder("abc");
-        try (OutputStream out = new OutputStreamTo(new File(folder, "A.txt"))) {
-            out.write("ABC".getBytes());
-        }
-        new File(folder, "B").mkdirs();
-        try (OutputStream out = new OutputStreamTo(
+        new File(folder, "A.txt").createNewFile();
+        new File(folder, "B").mkdir();
+        new File(
             new File(
-                folder,
-                new Joined(
-                    File.separator, "B", "B.txt"
-                ).asString()
-            )
-        )) {
-            out.write("EFG".getBytes());
-        }
-        new File(folder, "C").mkdirs();
+                new Joined(File.separator, folder.getPath(), "B").asString()
+            ), "B.txt"
+        ).createNewFile();
+        new File(folder, "C").mkdir();
         new Assertion<>(
             "Must zip directory with the same directory structure",
             new Sticky<>(
--- a/src/test/java/org/cactoos/io/ZipTest.java
+++ b/src/test/java/org/cactoos/io/ZipTest.java
@@ -26,12 +26,12 @@ package org.cactoos.io;
 
 import java.io.File;
 import java.io.OutputStream;
-import java.util.List;
 import java.util.stream.Collectors;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
 import org.cactoos.list.ListOf;
 import org.cactoos.scalar.LengthOf;
+import org.cactoos.scalar.Sticky;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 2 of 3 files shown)


0crat:
This pull request [#1321](https://github.com/yegor256/cactoos/issues/1321) is assigned to @victornoel[/z](https://www.0crat.com/u/victornoel), here is [why](https://www.0crat.com/footprint/C63314D6Z/4450eb39-9c42-4bd0-b28b-a946c274d926); the budget is 15 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§27](http://www.zerocracy.com/policy.html#27) and when you decide to accept the changes, inform @paulodamaso[/z](https://www.0crat.com/u/paulodamaso) (the architect) right in this ticket; if you decide that this PR should not be accepted ever, also inform the architect; this [blog post](http://www.yegor256.com/2015/02/09/serious-code-reviewer.html) will help you understand what is expected from a code reviewer; there will be no monetary reward for this job

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -67,7 +67,7 @@ public void zip() throws Exception {
                 "Can't list files in a directory represented by a path",
                 cnt,
                 // @checkstyle MagicNumber (1 line)
-                Matchers.equalTo(4)
+                Matchers.equalTo(1)
```
victornoel: @fanifieiev please convert this test to `Assertion` and OO matchers instead of static method. Don't forget about changing the assertion message to an assertive sentence ("Must..." instead of "Can't...").

victornoel: @fanifieiev Please replace this matcher with a direct comparison to an iterable of path as strings. (instead of counting them, let's compare them precisely).

victornoel:
@fanifieiev ping?

fanifieiev:
@victornoel Sorry for the delay. Please review again.

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,87 @@
...
+            new File(folder, "A.txt")
+        )
+        ) {
+            out.write("ABC".getBytes());
+        }
+        try (OutputStream out = new OutputStreamTo(
+            new File(folder, "B.txt")
+        )
+        ) {
+            out.write("EFG".getBytes());
```
victornoel: @fanifieiev is it really needed to have all this code to create a file? `new File(...).createNewfile()` is as good, no?

victornoel: @fanifieiev or even use `this.temporal.newFile` no?

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,87 @@
...
+        new Assertion<>(
+            "Must zip directory with the same directory structure",
+            new Sticky<>(
+                () -> {
+                    new LengthOf(
+                        new TeeInput(
+                            new InputOf(
+                                new Zip(
+                                    new Directory(folder)
+                                ).stream()
```
victornoel: @fanifieiev `Zip` is already an `Input`, no need to pass it to `InputOf`

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,87 @@
...
+                    ) {
+                        return file.stream().map(ZipEntry::toString).collect(
+                            Collectors.toList()
+                        );
+                    }
+                }
+            ),
+            new MatcherOf<>(
+                sclr -> {
+                    sclr.value().containsAll(
```
victornoel: @fanifieiev directly use `new IsEqual<>(new ListOf<>(..))` here

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,87 @@
...
+                        new ListOf<>(
+                            "abs\\A.txt",
```
victornoel: @fanifieiev is it really `abs`?

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -33,50 +33,58 @@
...
-    public Zip(final Directory origin) {
-        this.origin = origin;
+    public Zip(final Directory dir) {
+        this.directory = dir;
     }
 
     @Override
     public InputStream stream() throws Exception {
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
+        final Path parent = new ItemAt<>(0, this.directory).value().getParent();
```
victornoel: @fanifieiev what if it is empty? Please add a test to validate this will work. I believe there is some need to change the interface `Directory` to expose the `Path` itself...

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -33,50 +33,58 @@
...
+            new ForEach<Path>(
+                path -> {
+                    final File file = path.toFile();
+                    final String relative = parent.relativize(path).toString();
+                    if (file.isFile()) {
+                        zip.putNextEntry(new ZipEntry(relative));
+                        try (FileInputStream fis = new FileInputStream(file)) {
+                            zip.write(new BytesOf(new InputOf(fis)).asBytes());
+                        }
+                    } else if (!new Directory(file).iterator().hasNext()) {
```
victornoel: @fanifieiev what's happening here? it seems this does not go recursively into directories? Can you add a test to validate it works with sub directories?

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel It is not just creating file, it also puts some content into file

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel Good point

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel It won't work. `IsEqual` uses equals of List implementation, so it won't check the equality of each element

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel No, it is a mistake. And also I did missed `return` that's why the test did not fail. Will fix.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel It is never empty. If you pass `Directory `object, it has a path and it has a parent folder:

- For linux: the last parent would be `/`
- For win: the last parent would be `C` or `D`, etc
 Regarding the `I believe there is some need to change the interface Directory to expose the Path itself...`, yes it is required and I will leave a todo.

fanifieiev:
@victornoel Please review again

src/test/java/org/cactoos/io/ZipTest.java
victornoel: @fanifieiev I can see that, is this needed? I'm just bothered by the fact that the test contains more things that are not related to what is tested than things that are related to it :)

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -30,53 +30,68 @@
...
+                            new Directory(file).iterator()
+                        );
+                        if (!paths.hasNext()) {
+                            final org.cactoos.text.Joined join = new Joined(
+                                "/", relative, ""
+                            );
+                            zip.putNextEntry(new ZipEntry(join.toString()));
+                        }
                     }
+                    zip.closeEntry();
```
victornoel: @fanifieiev please wrap the code before in a try finally to ensure the entry is always properly closed

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -30,53 +30,68 @@
...
+                        try (FileInputStream fis = new FileInputStream(file)) {
+                            zip.write(new BytesOf(new InputOf(fis)).asBytes());
+                        }
+                    } else {
+                        final Iterator<Path> paths = new Skipped<>(
+                            1,
+                            new Directory(file).iterator()
+                        );
+                        if (!paths.hasNext()) {
+                            final org.cactoos.text.Joined join = new Joined(
```
victornoel: @fanifieiev why the fqdn for Joined? Best to use `Text` on the left side of the assignment anyway

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -30,53 +30,68 @@
...
+                                "/", relative, ""
+                            );
+                            zip.putNextEntry(new ZipEntry(join.toString()));
```
victornoel: @fanifieiev if possible please prefer `asString()` to invoke the behaviour of a `Text`

src/main/java/org/cactoos/io/Zip.java
victornoel: @fanifieiev you forgot that

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,87 @@
...
+                    ) {
+                        return new IterableOf<>(
+                            zipped.stream().map(
+                                e -> new TextOf(e.getName())
+                            ).collect(Collectors.toList())
+                        );
+                    }
+                }
+            ),
+            new ScalarHasValue<>(
```
victornoel: @fanifieiev can't we get rid of the `Scalar` stuffs in that test? It seems over-engineered

src/main/java/org/cactoos/io/Zip.java
victornoel: @fanifieiev can you add a test where the directory is empty at least?

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel Why not related??? Zipping(aka archiving) is the process to compress content of the files, so the content in files is related to zipping.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel Will do.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel it is outdated, there is a new code instead of that

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel will change

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel I tried, but this one was the most appropriate in this situation

src/test/java/org/cactoos/io/ZipTest.java
victornoel: @fanifieiev I'm sorry I'm not clear: do you test that the content is kept between zipping/unzipping? If yes, then it is related, if not then you can use empty file and the test will still be valid (and simpler to read). Let's keep tests as simple as possible.

src/main/java/org/cactoos/io/Zip.java
victornoel: @fanifieiev my comment is still valid, maybe I am not clear? I doubt when reading the code that it does recursively zip the directories, so instead of thinking too much, I'm asking you for a test that validates this expected behaviour.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel it is in `try `block already

victornoel: @fanifieiev let me be clearer: the content of the foreach loop should be in a try, and the `closeEntry` should be in a finally

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel ok, will use empty files

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel It does not do recursively. The `Directory` object provides the iterable of all the available paths.
The test contains subfolders.
So, the test has the following files and directories:
```
/abc/A.txt # file inside root folder
/abc/B/B.txt # file inside subfolder
/abc/C # empty subdirectory
```

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel Why? If the `closeEntry` fails there is no need to iterate over the rest of entries, so 
ZipOutputStream will be closed automatically.
https://docs.oracle.com/javase/7/docs/api/java/util/zip/ZipOutputStream.html#closeEntry()

victornoel: @fanifieiev ok, seems good to me then

src/main/java/org/cactoos/io/Zip.java
victornoel: @fanifieiev I missed that, ok. I think it would make sense to have multiple tests for the various aspects being tested and not everything in the same test though:
- empty dir
- multiple nested directories
- nominal behaviour with the correct base path being zipped

src/test/java/org/cactoos/io/ZipTest.java
victornoel: @fanifieiev ok, let's keep it like that, I don't see any way to make the test clearer except moving the zipping part out of the assertion argument.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel Agree, but I think we should have at least one test which would have complex case, when there is an empty folder, file in subfolder and just a file inside main folder.
I will also create a separate test for empty dir.

fanifieiev:
@victornoel Please review again.
I still do not like the way of creating directories and files.
Maybe we need to create some EO functionality to create files/dirs elegantly? WDYT?

victornoel:
@fanifieiev 
> I still do not like the way of creating directories and files.
> Maybe we need to create some EO functionality to create files/dirs elegantly? WDYT?

Maybe create an issue to discuss it?

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,113 @@
...
+                            ).asBytes()
+                        );
+                    }
+                    try (ZipFile zipped = new ZipFile(
+                        new File(folder.getParentFile(), "empty.zip")
+                    )
+                    ) {
+                        return zipped.stream().map(
+                            ZipEntry::toString
+                        ).findFirst().get();
```
victornoel: @fanifieiev can we get rid of `stream()` usage? Also this is not good, if there is a bug and there is two files in the zip, only the first one will be returned and the test will pass incorrectly

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,113 @@
...
+                        );
+                    }
+                    try (ZipFile zipped = new ZipFile(
+                        new File(folder.getParentFile(), zipname)
+                    )
+                    ) {
+                        return new IterableOf<>(
+                            zipped.stream().map(
+                                e -> new TextOf(e.getName())
+                            ).collect(Collectors.toList())
```
victornoel: @fanifieiev can we get rid of stream and collect here?

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel OK, makes sense

fanifieiev:
> @fanifieiev
> 
> > I still do not like the way of creating directories and files.
> > Maybe we need to create some EO functionality to create files/dirs elegantly? WDYT?
> 
> Maybe create an issue to discuss it?

@victornoel Will do

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel I am not sure how to do without `Stream`, I actually tried, but could not make it work.
If you look inside `stream()` method of `ZipEntry`, you will see a lot of things happening there.

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel Makes sense

src/test/java/org/cactoos/io/ZipTest.java
victornoel: @fanifieiev can't you wrap a stream into an `IterableOf` and then use `Mapped` on it for example? I think there should be something like that possible.

fanifieiev:
> > @fanifieiev
> > > I still do not like the way of creating directories and files.
> > > Maybe we need to create some EO functionality to create files/dirs elegantly? WDYT?
> > 
> > 
> > Maybe create an issue to discuss it?
> 
> @victornoel Will do

@victornoel I have created a ticket: https://github.com/yegor256/cactoos/issues/1333

fanifieiev:
@victornoel Please review again

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -75,14 +76,21 @@ public void mustZipEmptyDirectory() throws Exception {
                         new File(folder.getParentFile(), "empty.zip")
                     )
                     ) {
-                        return zipped.stream().map(
-                            ZipEntry::toString
-                        ).findFirst().get();
+                        return new IterableOf<>(
+                            new ListOf<>(
+                                new Mapped<>(
```
victornoel: @fanifieiev please use directly the iterable abstraction, you don't need list:
```
return new Mapped<>(
  ZipEntry::getName,
  zipped.stream()::iterator
)
```

should work

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -117,9 +125,12 @@ public void mustZipDirectory() throws Exception {
                     )
                     ) {
                         return new IterableOf<>(
-                            zipped.stream().map(
-                                e -> new TextOf(e.getName())
-                            ).collect(Collectors.toList())
+                            new ListOf<>(
+                                new Mapped<>(
+                                    entry -> new TextOf(entry.getName()),
+                                    zipped.stream().iterator()
```
victornoel: @fanifieiev same thing

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel I need ListOf, because I need to collect all the values into collection before I leave the try block, otherwise it fails to get the values.
I would not do like that, if I could use just as you posted.

victornoel: @fanifieiev in that case please remove the wrapping `IterableOf`, it is not needed (a `List` is an `Iterable`).

fanifieiev: @victornoel Right, but `IterableOf `has equals that can check the equality of each element, the `ListOf `does not have.

victornoel: @fanifieiev have you tried? `ListOf` extends `ListEnvelope` which delegate the `equals` method so it should work as expected IMHO or we should change the matcher so that it works.

fanifieiev: @victornoel Have you see the implementation of `equals` of `ListOf ` ???
The implementation of `equals `for `ListOf` is here https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/iterable/IterableEnvelope.java

victornoel: @fanifieiev yes, it delegates to the wrapped list, that's what I said

fanifieiev: @victornoel The warpped list in this case is `LinkedList`. It does not check the equality of each element.

fanifieiev: @victornoel Ok, will try to change the matcher

fanifieiev:
@victornoel Please review the latest changes

src/test/java/org/cactoos/io/ZipTest.java
victornoel: > The warpped list in this case is LinkedList. It does not check the equality of each element.

@fanifieiev yes it does… that's part of the contract of equals on all `List`… if you don't want to test for order, use a `SetOf` instead

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,121 @@
...
-    public final TemporaryFolder folder = new TemporaryFolder();
+    public final TemporaryFolder temporal = new TemporaryFolder();
+
+    @Test
+    public void mustZipEmptyDirectory() throws Exception {
+        final File folder = this.temporal.newFolder("empty");
+        new Assertion<>(
+            "Must zip empty directory",
+            new Sticky<>(
+                new IterableOf<>(
```
victornoel: @fanifieiev why do you add even more decorators? I propose the following:
- returns a `SetOf` inside the scalar, so that we don't bother about order
- use the matcher `IsEqual` with a `SetOf` there too

And it should work perfectly.

We don't want to use `HasValues` because it is not strict about the content, it only check that some things are present (like `contains()`)

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel You are right, AbstractList has equals implementation. I though it did not, because it did not work for me.

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel I need to wrap scalar lambda into iterable, that is why I am doing:
```
new IterableOf<>(
   () -> {...}
)
```

and I use `Sticky `because I need cached value.

victornoel: @fanifieiev I'm sorry, but that's way too complex. We should be able to simplify much more this code... what's the point of creating a list of a mapped iterator if after that you return the list's iterator and then you wrap it into an iterable that you make sticky? That's way too complex.

We spent already too much time here, please simplify that with as less as possible classes involved. Take the time to understand what's the expected behaviour of those classes and remove layers instead of adding more.

I see two simple solutions:
- the one you had before based on ScalarHasValue with a Sticky scalar responsible of holding the value and the lambda of the scalar just returning a `ListOf` or a `SetOf` (whichever, it's not important)
- move all this code in the lambda OUT of the assertion and simply store the resulting list or iterable or set or whatever in a variable.

victornoel: @fanifieiev I have created #1334 to explain what is the problem with `iterable.Sticky`.

fanifieiev: @victornoel OK

fanifieiev:
@victornoel Please take a look

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -30,53 +30,69 @@
...
+                            1,
+                            new Directory(file).iterator()
+                        );
+                        if (!paths.hasNext()) {
+                            zip.putNextEntry(
+                                new ZipEntry(
+                                    new Joined("/", relative, "").asString()
+                                )
+                            );
+                        }
```
victornoel: @fanifieiev one last thing: what is the purpose of this introduced else branch here? it wasn't there before and it doesn't seem to solve the problem from the original ticket you are solving.

fanifieiev: victornoel this else branch adds empty folder to zip file, if empty folder or subfolder is present.

victornoel: @fanifieiev it seems also complex for what it does, as a REV I'm not happy with it and I don't want to spend more time explaining what bothers me and what should be done.
So I think it would be safer to remove this and add a todo for this and let the work continue in another ticket/PR.

fanifieiev:
@victornoel I removed the functionality of zipping empty folders.Please take a look.

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -33,50 +33,54 @@
...
+                    final String relative = this.directory.path().getParent()
+                        .relativize(path).toString();
+                    final File file = path.toFile();
+                    if (file.isFile()) {
+                        zip.putNextEntry(new ZipEntry(relative));
+                        try (FileInputStream fis = new FileInputStream(file)) {
+                            zip.write(new BytesOf(new InputOf(fis)).asBytes());
+                        }
                     }
+                    zip.closeEntry();
```
victornoel: @fanifieiev this should be inside the `if` : the entry is created only if it is a file.

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,80 @@
...
-        }
+    public void mustZipDirectory() throws Exception {
+        final String zipname = "abc.zip";
+        final File folder = this.temporal.newFolder("abc");
+        new File(folder, "A.txt").createNewFile();
+        new File(folder, "B").mkdir();
+        new File(
+            new File(
+                new Joined(File.separator, folder.getPath(), "B").asString()
+            ), "B.txt"
```
victornoel: @fanifieiev this seems very complex, what is wrong with: `new File(new File(folder, "B"), "B.txt"))` ? You can even allocate `new File(folder, "B")` to a variable to be able to not repeat yourself

src/test/java/org/cactoos/io/ZipTest.java
```diff
@@ -24,51 +24,80 @@
...
+import org.llorllale.cactoos.matchers.ScalarHasValue;
 
 /**
  * Test case for {@link Zip}.
  *
  * @since 0.29
  * @checkstyle JavadocMethodCheck (500 lines)
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
  */
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
```
victornoel: @fanifieiev please remove this, it seems the only repeated literal is `abc`, `B.txt` and `A.txt`, so you should be able to allocate the `File` to a variable and refer to their name in the rest of the test.

src/main/java/org/cactoos/io/Zip.java
fanifieiev: @victornoel Agree

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel Agree

src/test/java/org/cactoos/io/ZipTest.java
fanifieiev: @victornoel Ok

fanifieiev:
@victornoel Please review

src/main/java/org/cactoos/io/Zip.java
```diff
@@ -33,50 +33,54 @@
...
     public InputStream stream() throws Exception {
         final ByteArrayOutputStream out = new ByteArrayOutputStream();
         try (ZipOutputStream zip = new ZipOutputStream(out)) {
-            for (final Path path : this.origin) {
-                final File file = path.toFile();
-                final ZipEntry entry = new ZipEntry(
-                    file.getPath()
-                );
-                zip.putNextEntry(entry);
-                if (file.isFile()) {
```
victornoel: @fanifieiev actually, now that I read this change in its greater context, I realise that the code before your changes was perfectly ok with the `closeEntry()` were it was because it did the `putNextEntry` entry BEFORE the `if`.

I think we should revert to the original behaviour and put back the test for empty directories, everything should be working without problem and there is no need for the `@todo`.

fanifieiev: @victornoel Ok, will revert 2 latest commits then

victornoel: @fanifieiev that's not what I said, I'm sorry I wasn't clear:
- you don't need the else branch
- you just need to put back the `putNextEntry` before the `if`
- and keep the `closeEntry` out of the `if` in that case.

In other words: you should revert back to the original code before any of your changes.

And from there:
- only add the relativisation of the path as required by the issue you are solving.
- if you want you can add the foreach for the loop, it makes sense
- then you can use the new tests you made and leave the new one for empty directories.
- if the test for empty directories does not pass, then you can add a todo

But please, do not try to tackle more, looking back, I think I should have refused the PR instead of taking so much of our time :( Sorry about that.

fanifieiev: @victornoel if I remove the else branch, the test for empty dir won't pass.
Actually, please refuse this PR, I will create another one later.
Thanks.

victornoel:
@paulodamaso let's refuse this, @fanifieiev will make another PR.
@fanifieiev again, it's too bad we took too long to realise this was going to eat too much of our energy, I'm not very clear as to what is the best strategy in such case, because I thought we could succeed ^^ My main feedback would be to change as less things as possible (relativise the path) while cleaning up stuffs in a purely syntactic way (using `ForEach` for example or `Assertion` in tests) and adding a test to validate the expected behaviour.

fanifieiev:
@paulodamaso I am closing this PR as it has already taken much time.
Will create another PR.

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @victornoel[/z](https://www.0crat.com/u/victornoel), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

0crat:
Code review was too long (15 days), architects (@paulodamaso) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

0crat:
Pull request [#1321](https://github.com/yegor256/cactoos/issues/1321) was not merged, no payment for `ARC`, see [§28](http://www.zerocracy.com/policy.html#28)

sereshqua:
@victornoel @fanifieiev this PR has the most amount of revisions i've seen in ZC :)

sereshqua:
@0crat quality good

fanifieiev:
> @victornoel @fanifieiev this PR has the most amount of revisions i've seen in ZC :)
@sereshqua So, @victornoel  is the winner and deserves double points :)

victornoel:
@0crat status

victornoel:
@fanifieiev I think it shows more a failure on my part than anything else :sweat_smile:

0crat:
> @0crat status ([here](https://github.com/yegor256/cactoos/issues/1321#issuecomment-605604881))

@victornoel This is what I know about this job in [C63314D6Z](https://www.0crat.com/p/C63314D6Z), as in [§32](http://www.zerocracy.com/policy.html#32):

  * The job [#1321](https://github.com/yegor256/cactoos/issues/1321) is not in scope
  * Job [footprint](https://www.0crat.com/footprint/C63314D6Z?q=%7Bjob%3A%27gh%3Ayegor256%2Fcactoos%231321%27%2Ctype%3A%7B%24not%3A%2F%5ENotify%2F%7D%7D) (restricted area)

sereshqua:
@0crat quality good

0crat:
> @0crat quality good ([here](https://github.com/yegor256/cactoos/issues/1321#issuecomment-605666585))

@sereshqua There is no quality review for [#1321](https://github.com/yegor256/cactoos/issues/1321), no performer

Issue #1287, opened 2020-02-06
TextEnvelope should be reimplemented to be a simple envelope


victornoel:
The current `TextEnvelope` implementation, like many of the other envelopes in cactoos, is plain wrong (see for example #1185 and other like that).

If `TextEnvelope` takes a `Scalar`, it is NOT an envelope, it's an implementation of `Text` based on a `Scalar`, and making it abstract is a mistake because 1) it promotes (bad) inheritance and 2) it can't be used directly.

I propose to:
- Move to `TextOf` the current implementation of `TextEnvelope`.
- make `TextEnvelope` just delegate `asString()`, `toString()`, `equals()` and `hashCode()` to a wrapped `Text`
- redefine all other `Text` that previously relied on `TextEnvelope` behaviour via inheritance by relying on  `TextOf` via wrapping by passing it to `TextEnvelope` as a superclass.

See also #942 #1156 for related tickets

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

victornoel:
@paulodamaso I think this should also be tackled soon, like with #1335. I just got bitten by it because `TextOf` nor `TextEnvelope` can be used when I need a `Text` from a `Scalar` while I should be able to do so.

paulodamaso:
@victornoel I see this "envleope" classes as default classes which have default methods for all instances; in this case, for example, it has the basic behavior for all `Text` instances (or should have). In theory their relation to the classes that extends them is not inheritance, but implementation / realization (according to UML theory); unfortunately java engineers decided that abstract classes have inheritance instead of realization relation to its subclasses, so in practice it have all the problems that we see in class inheritance. We could create default methods in `Text` interface with these behaviors, but I still don't know if it's a better solution. Just for the record, I don't like all the guys extending `TextEnvelope` either.

The problem I see is "envelope" seems a bad name to me. An envelope is something that wrap another thing, like an envelope and a letter; in this case we are using the name in the opposite of the real meaning: envelope is a basic / default / common implementation of `Text`, with methods that should be the same to every `Text` instance.

I'm still a bit unsure how we could solve these problems; seems that changing the code as you propose don't avoid the problem of getting wrong results on `Text` implementations that don't have the `ScalarText` wrapped inside them, for example.

wladyan:
[...]

Issue #1241, opened 2019-11-16
A new kind of Equals?


fabriciofx:
I see a lot of this (bellow) code repeat over and over:
```
@Override
public boolean equals(final Object other) {
    return new Unchecked<>(
        new Or(
            () -> other == this,
            new And(
                () -> other != null,
                () -> Collection.class.isAssignableFrom(other.getClass()),
                () -> {
                    final Collection<?> compared = (Collection<?>) other;
                    return this.size() == compared.size();
                },
                () -> {
                    final Collection<?> compared = (Collection<?>) other;
                    final Iterator<?> iterator = compared.iterator();
                    return new Unchecked<>(
                        new And(
                            (T input) -> input.equals(iterator.next()),
                            this
                        )
                    ).value();
                }
            )
        )
    ).value();
}
```
and I think is possible encapsulate it in one class to be reused. WDYT?

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

fabriciofx:
@victornoel WDYT?

victornoel:
@fabriciofx so something similar to `HashCode` but for `Equals`? What I am worried about is that this code should not be duplicated so much normally. Each of the classes that implements equals should have their own, internal state related, implement IMHO.

Could you make a list of the classes that would use this `Equals` and see in practice what it would entails?

victornoel:
@fabriciofx following some discussion with @andreoss (see https://github.com/yegor256/cactoos/pull/1458#discussion_r487550572), I now realise that maybe something we could do is provide, for each of the interfaces supported by cactoos (and for which it makes sense), a class `Equals implements Scalar<Boolean>` and a class `HashCode implements Scalar<Integer>` that the implementations of those interfaces could reuse when needed.

@skapral I remember you had some opinion about equals and hashCode, don't hesitate to chime in :)

There is also a debate in #1156 where @fanifieiev (don't hesitate to chime in too :) made an interesting comment (on top of all the others there): https://github.com/yegor256/cactoos/issues/1156#issuecomment-553619095

There is also #920 with some similar concerns for the proc/func/scalar packages but I don't think those are classes that would really benefit of having equals IMHO because they don't really have a comparable state exposed through their behaviour.

Issue #1156, opened 2019-07-02
TextEnvelope equals is not symmetric 


g4s8:
https://github.com/yegor256/cactoos/blob/100885a4638555d30fb56beefbb6691108b30d0f/src/main/java/org/cactoos/text/TextEnvelope.java#L79-L90

From Java `Object` [docs](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#equals(java.lang.Object)):
>  The equals method implements an equivalence relation on non-null object references:
>
>    It is reflexive: for any non-null reference value x, x.equals(x) should return true.
>    It is symmetric: for any non-null reference values x and y, x.equals(y) should return true if  and only if y.equals(x) returns true.
>    It is transitive: for any non-null reference values x, y, and z, if x.equals(y) returns true and y.equals(z) returns true, then x.equals(z) should return true.
>    It is consistent: for any non-null reference values x and y, multiple invocations of x.equals(y) consistently return true or consistently return false, provided no information used in equals comparisons on the objects is modified.
>    For any non-null reference value x, x.equals(null) should return false. 

But `TextEnvelope` `equals()` implementation is not symmetric:
```java
class Env extends TextEnvelope {
  public Env(String str) {
    super(new TextOf(str));
  }
}

class Txt implements Text {
  private final String str;
  public Txt(final String str) {
    this.str = str;
  }
  @Override
  public String asString() {
    return this.str;
  }
}

final String str = "test";
Text env = new Env(str)
Text txt = new Txt(str)
env.equals(txt) // returns true
txt.equals(env) // returns false
```

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

0crat:
@g4s8[/z](https://www.0crat.com/u/g4s8) this project will fix the problem faster if you donate a few dollars to it; just [click here](https://www.0crat.com/contrib/C63314D6Z) and pay via Stripe, it's very fast, convenient and appreciated; thanks a lot!

llorllale:
@g4s8 you rolled your own implementation of `Text` with `Txt` in your example... that means you didn't extend `TextEnvelope`, and I see that you also didn't implement `equals()`. Hence this happens: `txt.equals(env) // returns false`

g4s8:
@llorllale that's true, but I think it's a valid case, since anybody can implement `Text` interface without `equals()` or with standard `equals()` implementation, e.g.
```java
[...]

PR #1458, opened 2020-09-13
(#1287) TextEnvelope is a real envelope

victornoel:
This is for #1287.

### DIFF
--- a/src/main/java/org/cactoos/text/Synced.java
+++ b/src/main/java/org/cactoos/text/Synced.java
@@ -23,7 +23,6 @@
  */
 package org.cactoos.text;
 
-import org.cactoos.Scalar;
 import org.cactoos.Text;
 
 /**
@@ -48,13 +47,6 @@ public final class Synced extends TextEnvelope {
      * @param lck The lock
      */
     public Synced(final Text text, final Object lck) {
-        super(new Scalar<String>() {
-            @Override
-            public String value() throws Exception {
-                synchronized (lck) {
-                    return text.asString();
-                }
-            }
-        });
+        super(new TextOf(new org.cactoos.scalar.Synced<>(text::asString, lck)));
     }
 }
--- a/src/main/java/org/cactoos/text/TextEnvelope.java
+++ b/src/main/java/org/cactoos/text/TextEnvelope.java
@@ -23,71 +23,46 @@
  */
 package org.cactoos.text;
 
-import java.io.IOException;
-import org.cactoos.Scalar;
 import org.cactoos.Text;
-import org.cactoos.scalar.And;
-import org.cactoos.scalar.IoChecked;
-import org.cactoos.scalar.Or;
-import org.cactoos.scalar.Unchecked;
 
 /**
- * Text envelope that provides {@link #equals(Object)} and {@link #hashCode()}
- * methods.
+ * {@link Text} envelope.
+ *
  * @since 0.32
  * @checkstyle AbstractClassNameCheck (500 lines)
  */
 public abstract class TextEnvelope implements Text {
 
     /**
-     * String value of the envelope.
+     * Wrapped Text.
      */
-    private final IoChecked<String> origin;
+    private final Text origin;
 
     /**
      * Ctor.
      * @param text Text representing the text value.
      */
     public TextEnvelope(final Text text) {
-        this(new IoChecked<>(text::asString));
-    }
-
-    /**
-     * Ctor.
-     * @param scalar Scalar representing the text value.
-     */
-    public TextEnvelope(final Scalar<String> scalar) {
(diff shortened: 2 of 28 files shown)


0crat:
Job [#1458](https://github.com/yegor256/cactoos/issues/1458) doesn't exist, can't set role

victornoel:
@0crat in

src/main/java/org/cactoos/text/Synced.java
```diff
@@ -48,13 +47,14 @@ public Synced(final Text text) {
...
      */
     public Synced(final Text text, final Object lck) {
-        super(new Scalar<String>() {
-            @Override
-            public String value() throws Exception {
-                synchronized (lck) {
-                    return text.asString();
+        super(
+            new TextOf(
+                () -> {
```
andreoss, line 52: @victornoel Can we reused `scalar.Synched` here?

src/main/java/org/cactoos/text/TextEnvelope.java
```diff
@@ -23,71 +23,46 @@
...
     public final String toString() {
-        return new UncheckedText(this).asString();
+        return this.origin.toString();
     }
 
     @Override
-    public final int hashCode() {
-        return new Unchecked<>(this.origin).value().hashCode();
+    public final boolean equals(final Object obj) {
+        return this.origin.equals(obj);
```
andreoss, line 61: @victornoel Wouldn't this definition of `equals` cause problems? The operation seems to loose symmetric properties. 
For example
`new TextEnvelope(x).equals(x) == true`, but  not necessarily `x.equals(new TextEnvelope(x))`

victornoel:
@0crat status

src/main/java/org/cactoos/text/TextEnvelope.java
victornoel: @andreoss this is a known limitation of `equals` actually when an interface has many different implementations. This subject has been discussed many times in the EO community, i.e., what does method equals mean for objects (you can search for "equals" in cactoos issues for example).

My take on it is that if you start having multiple implementations of the same interface, then you can't just expect equals to rely on some internal state, and thus, when you implement it, you have to rely on the exposed behaviour of the interface.

In the case of `Text`, it means that all implementations of `Text` should rely on `Text.asString` to implement `equals`, so in that case, the symmetry will be conserved. It is the same for `MapOf`, `IterableOf` and the other collection-related classes for example. Also sometimes it's better not to implement `equals` at all (e.g., `Iterator`).

victornoel: @andreoss see for example https://github.com/yegor256/cactoos/issues/1156#issuecomment-511189091 where I made the same kind of comment. Also I started a discussion on all that matter in #1241, don't hesitate to chime in.

0crat:
> @0crat status ([here](https://github.com/yegor256/cactoos/issues/1458#issuecomment-691693814))

@victornoel This is what I know about this job in C63314D6Z, as in [§32](http://www.zerocracy.com/policy.html#32):

  * The job [#1458](https://github.com/yegor256/cactoos/issues/1458) is [in scope](http://www.zerocracy.com/policy.html#14) for 2hr
  * The role is `REV`
  * The job is assigned to @andreoss[/z](https://www.0crat.com/u/andreoss) for [1hr](http://www.zerocracy.com/policy.html#8)
  * There is no monetary reward attached, it's a [free](http://www.zerocracy.com/policy.html#2) job
  * The job doesn't have any [impediments](http://www.zerocracy.com/policy.html#9)
  * The [budget](http://www.zerocracy.com/policy.html#4) is 15 minutes/points
  * These users are banned and won't be assigned:
    * @victornoel[/z](https://www.0crat.com/u/victornoel): This user reported the ticket
  * Job [footprint](https://www.0crat.com/footprint/C63314D6Z?q=%7Bjob%3A%27gh%3Ayegor256%2Fcactoos%231458%27%2Ctype%3A%7B%24not%3A%2F%5ENotify%2F%7D%7D) (restricted area)

src/main/java/org/cactoos/text/Synced.java
victornoel: @andreoss good catch

victornoel:
@andreoss see my changes and answers

andreoss (approved):
@victornoel Looks fine

victornoel:
@rultor merge

0crat:
Job was finished in 3 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

0crat:
Job audit: performer @andreoss[/z](https://www.0crat.com/u/andreoss)[/z](https://www.0crat.com/u/andreoss) didn't make CR comments

Issue #1384, opened 2020-05-09
MapEnvelope should be a real envelope and the behaviour moved to MapOf


victornoel:
In the same way we improved the cactoos API by making `ListEnvelope`, `SetEnvelope` and so on real envelopes with actual behaviour such as `equals`, `toString`, `hashcode`, etc, moved to the `ListOf`, `SetOf`, etc classes. I believe we should continue and do the same for `MapEnvelope` and `MapOf`.

The final objective is to have `MapOf` as the main entry point to build instances of `Map` in cactoos (based on a `HashMap`), and have `Sticky` and `Solid` removed.

The existing `Grouped` and `Merged` classes seems to be coherent with this desired design, so that's good.

0crat:
@paulodamaso[/z](https://www.0crat.com/u/paulodamaso) please, pay attention to this issue

victornoel:
@0crat in

victornoel:
@0crat assign @andreoss

victornoel:
@andreoss I'm starting to assign manually because if not 0crat always assign latest issues and I would like to advance on those. I hope it's ok with you :) no rush anyway

victornoel:
@andreoss thx :)

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @andreoss[/z](https://www.0crat.com/u/andreoss), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality good

PR #1463, opened 2020-09-15
#1384: Remove lazy initizaliaztion from MapEnvelope

andreoss:
Per #1384 
- Make MapEnvelope a proper envelope
- Add ctor for `MapOf`
- Remove tests for lazy initialization


fabriciofx:
@andreoss something goes wrong. Can you take a look?

andreoss:
@fabriciofx Done. Ready for review.

fabriciofx:
@andreoss @victornoel I'm sorry, but I didn't figure out why we should remove lazy initialization for `MapEnvelope`. Can explain it, please?

andreoss:
@fabriciofx @victornoel 
As I understand, a proper envelope should just delegate to a wrapped object and nothing else. 
Lazy initialization cannot be done inside `MapEnvelope`, we can introduce a new class `map.Delayed`, and bring back those two tests I removed here. 
I don't like the fact that map is being created inside constructor, but it appears there is no other solution without extra implementation of `Map`

fabriciofx:
> I don't like the fact that map is being created inside constructor
I didn't like it too, that's why I'm asking for it. @victornoel WDYT?

andreoss:
@fabriciofx BTW, the same happens in 
https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/list/ListOf.java#L65
https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/set/SetOf.java#L56

fabriciofx:
@andreoss hummm. I particularly don't like this solution. Collections can have a lot of elements and add them takes a lot of time. @victornoel WDYT?

src/main/java/org/cactoos/map/Solid.java
```diff
@@ -148,8 +148,10 @@ public Solid(final Map<X, Y> map,
      */
     public Solid(final Map<X, Y> map) {
```
victornoel, line 149: @andreoss this class should be removed too as discussed in the ticket

src/main/java/org/cactoos/map/Sticky.java
```diff
@@ -158,12 +158,14 @@ public Sticky(final Map<X, Y> map,
      */
     public Sticky(final Map<X, Y> map) {
```
victornoel, line 159: @andreoss this class should be removed too as discussed in the ticket

src/main/java/org/cactoos/map/Synced.java
```diff
@@ -150,12 +150,14 @@ public Synced(final Map<X, Y> map,
...
     public Synced(final Map<X, Y> map) {
         super(
-            new org.cactoos.scalar.Synced<>(
-                () -> {
-                    final Map<X, Y> temp = new ConcurrentHashMap<>(0);
-                    temp.putAll(map);
-                    return temp;
-                }
+            new MapOf<>(
+                new org.cactoos.scalar.Synced<>(
```
victornoel, line 154: @andreoss there is no need for wrapping that into Synced anymore since it is evaluated in the constructor

src/test/java/org/cactoos/map/MapOfTest.java
```diff
@@ -78,24 +74,6 @@ void convertsIterableToMap() {
         );
     }
 
-    @Test
-    void sensesChangesInMap() throws Exception {
```
victornoel, line 82: @andreoss what is the problem with this test?

src/test/java/org/cactoos/scalar/PropertiesOfTest.java
```diff
@@ -107,24 +102,4 @@ void convertsMapToProperties() {
         ).affirm();
     }
 
-    @Test
-    void sensesChangesInMap() throws Exception {
```
victornoel, line 111: @andreoss here too

src/main/java/org/cactoos/map/MapOf.java
```diff
@@ -148,12 +150,20 @@ public MapOf(final Map<X, Y> src,
...
     public MapOf(final Iterable<Map.Entry<X, Y>> entries) {
-        super(() -> {
+        this(() -> {
             final Map<X, Y> temp = new HashMap<>(0);
             for (final Map.Entry<X, Y> entry : entries) {
                 temp.put(entry.getKey(), entry.getValue());
             }
             return Collections.unmodifiableMap(temp);
         });
     }
```
victornoel, line 160: @andreoss instead of using a Scalar based constructor, I think it would make more sense to follow the same pattern as in `ListOf` and `SetOf`.

victornoel (changes requested):
@fabriciofx I believe the objective is quite clear from the ticket: `MapOf` should represent a concrete map, like `ListOf` and `SetOf` do. Cactoos is meant to implement OO interfaces, and List, Set, Map or even Collection are not OO interfaces, they are collection interfaces. So it was decided to only introduce the few classes needed to create new list, map, set, etc using constructors, but everything else related to lazy evaluation of an iterable object should go in iterable decorators.

@andreoss I think this is a good start but it's not complete, see my comments :)

src/test/java/org/cactoos/scalar/PropertiesOfTest.java
andreoss: @victornoel This test (and the one above) checks either map is lazily created or not, thus it's removed

src/main/java/org/cactoos/map/Synced.java
andreoss: @victornoel Then, there's no point in `map.Synced`. Should it be removed?

src/test/java/org/cactoos/scalar/PropertiesOfTest.java
victornoel: @andreoss ok, anyway, those tests are so strange that even if they had sense they would be unmaintainable ^^ thx :)

src/main/java/org/cactoos/map/Synced.java
```diff
@@ -150,12 +150,14 @@ public Synced(final Map<X, Y> map,
...
+                    () -> {
+                        final Map<X, Y> temp = new ConcurrentHashMap<>(0);
```
victornoel, line 156: @andreoss here please replace all of this by just wrapping the map into `Collections.synchronizedMap` as with `list.Synced`

src/main/java/org/cactoos/map/Sticky.java
andreoss: @victornoel Done.

src/main/java/org/cactoos/map/Synced.java
andreoss: @victornoel Replaced with `Collections.synchronizedMap`

src/main/java/org/cactoos/map/Synced.java
andreoss: @victornoel Done.

fabriciofx:
@andreoss I think all comments I can do it in this PR already done by @victornoel. @victornoel can you merge it, please?

src/main/java/org/cactoos/map/Grouped.java
```diff
@@ -57,16 +57,19 @@ public Grouped(
...
-            );
-            return stream.collect(
-                Collectors.groupingBy(
-                    keys,
-                    Collectors.mapping(values, Collectors.toList())
-                )
-            );
-        });
+        super(
+            new MapOf<>(
```
victornoel, line 61: @andreoss normally you don't need to go through `MapOf`, the `stream.collect` returns a `Map` that can be passed directly to `super`. Right?

src/main/java/org/cactoos/map/MapOf.java
```diff
@@ -143,17 +140,22 @@ public MapOf(final Map<X, Y> src,
         );
     }
 
+    /**
+     * Ctor.
+     * @param scalar The scalar for Map.
+     */
+    public MapOf(final Scalar<Map<X, Y>> scalar) {
+        this(new Unchecked<>(scalar).value().entrySet());
```
victornoel, line 148: @andreoss what is the `Scalar` for? If we want to create a map from an existing one, then let's just take a map in the constructor. You can then use `putAll` to add all the elements of the original map.

victornoel (changes requested):
@andreoss @fabriciofx actually there are more things to fix first :)

src/main/java/org/cactoos/map/Grouped.java
andreoss: @victornoel Done

src/main/java/org/cactoos/map/MapOf.java
andreoss: @victornoel Removed that ctor.

src/main/java/org/cactoos/map/Merged.java
```diff
@@ -42,11 +42,12 @@
...
-            new ListOf<>(maps)
-                .stream()
-                .flatMap(map -> map.entrySet().stream())
-                .collect(Collectors.toList())
+        super(
+            new MapOf<>(
+                new ListOf<>(maps)
+                    .stream()
+                    .flatMap(map -> map.entrySet().stream())
+                    .collect(Collectors.toList())
```
victornoel, line 50: @andreoss can't we manage to avoid creating an intermediate list just to iterate over it in MapOf? I'm not very clear about what streams provide us for that though...

victornoel: @andreoss well actually, this is out of scope of this PR, let's forget about it, I will create an issue.

victornoel:
@andreoss thx

victornoel:
@rultor merge

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @fabriciofx[/z](https://www.0crat.com/u/fabriciofx), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@fabriciofx please make sure you start all your comments with the name of the user they are addressed  to: [see](https://github.com/yegor256/cactoos/pull/1463#issuecomment-692787214)

sereshqua:
@0crat quality acceptable

Issue #1460, opened 2020-09-13
Abbreviated.java:83-87: Introduce `text.Flatten` that...


0pdd:
The puzzle `1287-014a5155` from #1287 has to be resolved: 

https://github.com/yegor256/cactoos/blob/47aff6e4c70f6f53d24bdccb7c1f6c12abfe60d2/src/main/java/org/cactoos/text/Abbreviated.java#L83-L87

The puzzle was created by Victor Noël on 13-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
This puzzle inherited milestone `1.0` from issue #1287.

victornoel:
@0crat assign me

victornoel:
@0crat wait for #1475 to be done first.

victornoel:
@0crat wait for #1515

0crat:
> @0crat wait for #1515 ([here](https://github.com/yegor256/cactoos/issues/1460#issuecomment-751481751))

@victornoel Job [#1460](https://github.com/yegor256/cactoos/issues/1460) is already on hold

andreoss:
@victornoel It's not quite clear why calling `.value()` directly in order to get `Text` is undesirable.

victornoel:
@andreoss I don't understand the question, can you clarify?

0pdd:
The puzzle `1287-014a5155` has disappeared from the source code, that's why I closed this issue.

victornoel:
@0crat status

0crat:
> @0crat status ([here](https://github.com/yegor256/cactoos/issues/1460#issuecomment-757498509))

@victornoel This is what I know about this job in C63314D6Z, as in [§32](http://www.zerocracy.com/policy.html#32):

  * The job [#1460](https://github.com/yegor256/cactoos/issues/1460) is not in scope
  * Job [footprint](https://www.0crat.com/footprint/C63314D6Z?q=%7Bjob%3A%27gh%3Ayegor256%2Fcactoos%231460%27%2Ctype%3A%7B%24not%3A%2F%5ENotify%2F%7D%7D) (restricted area)

PR #1465, opened 2020-09-18
Added Capitalized

kokodyn:
Added EO version of StringUtils.capitalize

I hope for your understanding, this is my first pull request to open source project.

### DIFF
--- a/src/main/java/org/cactoos/text/Capitalized.java
+++ b/src/main/java/org/cactoos/text/Capitalized.java
@@ -28,8 +28,9 @@ import org.cactoos.Text;
 import org.cactoos.scalar.Ternary;
 
 /**
- * Text in capitalized case, changed the first character to title case, no other
- * characters are changed.
+ * Text in capitalized case,
+ * changed the first character to title case as per {@link Character#toTitleCase(int)},
+ * no other characters are changed.
  *
  * @since 0.46
  */
@@ -53,25 +54,32 @@ public final class Capitalized extends TextEnvelope {
         super(
             new TextOf(
                 () -> {
-                    final Func<String, String> capitalize = cap -> {
-                        final Text first = new Sub(cap, 0, 1);
-                        final Text upper = new Upper(first);
+                    final Func<Text, Text> capitalize = cap -> {
+                        final Text titled = new TextOf(
+                            Character.toChars(
+                                Character.toTitleCase(
+                                    cap.asString().codePointAt(0)
+                                    )
+                                )
+                            );
                         return
-                            new Ternary<String>(
-                                () -> upper.equals(first),
+                            new Ternary<Text>(
+                                () -> new StartsWith(cap, titled).value(),
                                 () -> cap,
-                                () -> new StringBuilder()
-                                    .append(upper)
+                                () -> new TextOf(new StringBuilder()
+                                    .append(titled.asString())
                                     .append(new Sub(text, 1))
-                                    .toString())
+                                    .toString()
+                                    )
+                                )
                             .value();
                     };
-                    return new Ternary<String>(
-                        text.asString(),
-                        t -> t.isEmpty(),
+                    return new Ternary<Text>(
+                        text,
+                        t -> new IsBlank(t).value(),
                         t -> t,
                         t -> capitalize.apply(t)
-                            ).value();
+                            ).value().asString();
                 }));
     }
 }
--- a/src/main/java/org/cactoos/text/Capitalized.java
+++ b/src/main/java/org/cactoos/text/Capitalized.java
@@ -23,7 +23,9 @@
  */
 package org.cactoos.text;
 
+import org.cactoos.Func;
 import org.cactoos.Text;
+import org.cactoos.scalar.Ternary;
 
 /**
  * Text in capitalized case, changed the first character to title case, no other
@@ -31,7 +33,7 @@ import org.cactoos.Text;
  *
  * @since 0.46
  */
-public class Capitalized extends TextEnvelope {
+public final class Capitalized extends TextEnvelope {
 
     /**
      * Ctor.
@@ -51,16 +53,25 @@ public class Capitalized extends TextEnvelope {
         super(
             new TextOf(
                 () -> {
-                    String capitalized = text.asString();
-                    if (!capitalized.isEmpty()) {
-                        final Text first = new Sub(capitalized, 0, 1);
+                    final Func<String, String> capitalize = cap -> {
+                        final Text first = new Sub(cap, 0, 1);
                         final Text upper = new Upper(first);
-                        if (!upper.equals(first)) {
-                            capitalized = new StringBuilder().append(upper).append(new Sub(text, 1))
-                                .toString();
-                        }
-                    }
-                    return capitalized;
+                        return
+                            new Ternary<String>(
+                                () -> upper.equals(first),
+                                () -> cap,
+                                () -> new StringBuilder()
+                                    .append(upper)
+                                    .append(new Sub(text, 1))
+                                    .toString())
+                            .value();
+                    };
+                    return new Ternary<String>(
+                        text.asString(),
+                        t -> t.isEmpty(),
+                        t -> t,
+                        t -> capitalize.apply(t)
+                            ).value();
                 }));
     }
 }
--- a/src/main/java/org/cactoos/text/Capitalized.java
+++ b/src/main/java/org/cactoos/text/Capitalized.java
@@ -46,7 +46,7 @@ public final class Capitalized extends TextEnvelope {
 
     /**
      * Ctor.
-     * TODO replace Upper by Titled
+     * TODO replace {@link Character#toTitleCase(int)} by it's EO version
 
      * @param text The text
      */
@@ -59,7 +59,13 @@ public final class Capitalized extends TextEnvelope {
                         () -> text,
                         () -> new Joined(
                             "",
-                            new Upper(new Sub(text, 0, 1)),
+                            new TextOf(
+                                Character.toChars(
+                                    Character.toTitleCase(
+                                        new Sub(text, 0, 1).asString().codePointAt(0)
+                                    )
+                                )
+                            ),
                             new Sub(text, 1)
                             )
                     ).value().asString();
--- a/src/main/java/org/cactoos/text/Capitalized.java
+++ b/src/main/java/org/cactoos/text/Capitalized.java
@@ -23,7 +23,6 @@
  */
 package org.cactoos.text;
 
-import org.cactoos.Func;
 import org.cactoos.Text;
 import org.cactoos.scalar.Ternary;
 
@@ -47,37 +46,22 @@ public final class Capitalized extends TextEnvelope {
 
     /**
      * Ctor.
-     *
+     * TODO replace Upper by Titled
+
      * @param text The text
      */
     public Capitalized(final Text text) {
         super(
             new TextOf(
                 () -> {
-                    final Func<Text, Text> capitalize = cap -> {
-                        final Text titled = new TextOf(
-                            Character.toChars(
-                                Character.toTitleCase(
-                                    cap.asString().codePointAt(0)
-                                )
-                            )
-                        );
-                        return
-                            new Ternary<Text>(
-                                new StartsWith(cap, titled),
-                                () -> cap,
-                                () -> new Joined(
-                                    "",
-                                    titled,
-                                    new Sub(text, 1)
-                                )
-                            )
-                            .value();
-                    };
                     return new Ternary<Text>(
                         new IsBlank(text),
                         () -> text,
-                        () -> capitalize.apply(text)
+                        () -> new Joined(
+                            "",
+                            new Upper(new Sub(text, 0, 1)),
+                            new Sub(text, 1)
+                            )
                     ).value().asString();
                 }));
     }
--- a/src/main/java/org/cactoos/text/Capitalized.java
+++ b/src/main/java/org/cactoos/text/Capitalized.java
@@ -59,27 +59,26 @@ public final class Capitalized extends TextEnvelope {
                             Character.toChars(
                                 Character.toTitleCase(
                                     cap.asString().codePointAt(0)
-                                    )
                                 )
-                            );
+                            )
+                        );
                         return
                             new Ternary<Text>(
-                                () -> new StartsWith(cap, titled).value(),
+                                new StartsWith(cap, titled),
                                 () -> cap,
-                                () -> new TextOf(new StringBuilder()
-                                    .append(titled.asString())
-                                    .append(new Sub(text, 1))
-                                    .toString()
-                                    )
+                                () -> new Joined(
+                                    "",
+                                    titled,
+                                    new Sub(text, 1)
                                 )
+                            )
                             .value();
                     };
                     return new Ternary<Text>(
-                        text,
-                        t -> new IsBlank(t).value(),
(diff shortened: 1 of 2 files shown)


src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * Ctor.
+     *
+     * @param text The text
+     */
+    public Capitalized(final Text text) {
+        super(
+            new TextOf(
+                () -> {
+                    String capitalized = text.asString();
```
andreoss, line 54: @kokodyn Let's remove mutable variable from here. I guess `Ternary` can be used

src/test/java/org/cactoos/text/CapitalizedTest.java
```diff
@@ -0,0 +1,93 @@
...
+public class CapitalizedTest {
+    /**
+     * Text starting with upper case character.
+     */
+    private static final String UPPER_CASE = "Abc";
+
+    @Test
+    public void capitalizeEmptyText() {
+        new Assertion<>(
+            "Can't capitalize an empty text",
```
andreoss, line 43: @kokodyn I suggest `must` instead of `cannot`

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,66 @@
...
+
+import org.cactoos.Text;
+
+/**
+ * Text in capitalized case, changed the first character to title case, no other
+ * characters are changed.
+ *
+ * @since 0.46
+ */
+public class Capitalized extends TextEnvelope {
```
andreoss, line 34: @kokodyn May be we should name it `Title` instead?
Should it make all non-first characters lower-cased?

andreoss, line 34: @kokodyn The class should be `final`

src/test/java/org/cactoos/text/CapitalizedTest.java
```diff
@@ -0,0 +1,93 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.text;
+
+import org.junit.Test;
```
andreoss, line 26: @kokodyn Please use `Junit 5` and make class `final`, but not `public`

src/test/java/org/cactoos/text/CapitalizedTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * Test case for {@link Capitalized}.
+ * @since 0.46
+ */
+public class CapitalizedTest {
+    /**
+     * Text starting with upper case character.
+     */
+    private static final String UPPER_CASE = "Abc";
```
andreoss, line 38: @kokodyn I believe this constant is not necessary. If it causes Qulice complain, it would be better to suppress that.

andreoss (changes requested):
@kokodyn Please see my comments.
@victornoel There's no issue attached to this PR, should we create one?

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: Done but I am not fully convinced to using Ternary. It seems be more complicated than previous version.

src/test/java/org/cactoos/text/CapitalizedTest.java
kokodyn: I prefer to use "Can't" because as I see it it used commonly in other tests in similar context.

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: I will create ticket for "Capitalized" issue and we can discuss further,

kokodyn: done

src/test/java/org/cactoos/text/CapitalizedTest.java
kokodyn: I hope I did it in proper way, I need to update path to @Test annotation to allow Eclipse to handle it.

src/test/java/org/cactoos/text/CapitalizedTest.java
kokodyn: done

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: I created issue for this:
https://github.com/yegor256/cactoos/issues/1466

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Ctor.
+     *
+     * @param text The text
+     */
+    public Capitalized(final Text text) {
+        super(
+            new TextOf(
+                () -> {
+                    final Func<String, String> capitalize = cap -> {
```
andreoss, line 56: @kokodyn 
I believe it should be more like:
```
super(new Ternary<>(new IsBlank(text), () -> text, new Ternary<>(new StartsWith(...), ...);
```
Also, `Func<String, String>` should be a `Text` but without extra checks

src/test/java/org/cactoos/text/CapitalizedTest.java
andreoss: @kokodyn Ok, I didn't notice, that "can't" is so common

andreoss (changes requested):
@kokodyn There's more

src/test/java/org/cactoos/text/CapitalizedTest.java
victornoel: @kokodyn please let's use must as it is the default we use in cactoos (there are still examples where can't is used but this should be changed bit by bit)

src/test/java/org/cactoos/text/CapitalizedTest.java
```diff
@@ -68,8 +65,8 @@ public void capitalizeSingleUpperCaseText() {
     public void capitalizeTextStartingWithUpperCaseCharacter() {
```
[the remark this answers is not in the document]
kokodyn: done

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: Ok. Not sure what does it mean "without extra checks". Please look at my proposition in newest commit.

src/test/java/org/cactoos/text/CapitalizedTest.java
kokodyn: done

src/main/java/org/cactoos/text/Capitalized.java
andreoss: @kokodyn It looks complicated. 
Why it cannot be like that?
```
                new Ternary<Text>(
                    new IsBlank(text),
                    text,
                    new Joined(
                        "",
                        new Upper(new Sub(text, 0, 1)),
                        new Sub(text, 1)
                    )
                )
```

May be imperative approach was better here.

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,85 @@
...
+                                    .append(titled.asString())
+                                    .append(new Sub(text, 1))
+                                    .toString()
+                                    )
+                                )
+                            .value();
+                    };
+                    return new Ternary<Text>(
+                        text,
+                        t -> new IsBlank(t).value(),
```
victornoel, line 79: @kokodyn here too you can directly pass `IsBlank` instead of a lambda

victornoel (changes requested):
@kokodyn here are some comments :) all of this seems a bit complex for now, let's see if we can simplify it bit by bit!

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,85 @@
...
+                        final Text titled = new TextOf(
+                            Character.toChars(
+                                Character.toTitleCase(
+                                    cap.asString().codePointAt(0)
+                                    )
+                                )
+                            );
+                        return
+                            new Ternary<Text>(
+                                () -> new StartsWith(cap, titled).value(),
```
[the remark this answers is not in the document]
kokodyn: @victornoel done

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,85 @@
...
+            new TextOf(
+                () -> {
+                    final Func<Text, Text> capitalize = cap -> {
+                        final Text titled = new TextOf(
+                            Character.toChars(
+                                Character.toTitleCase(
+                                    cap.asString().codePointAt(0)
+                                    )
+                                )
+                            );
```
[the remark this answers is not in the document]
kokodyn: @victornoel I hope now it looks better.

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,85 @@
...
+                        return
+                            new Ternary<Text>(
+                                () -> new StartsWith(cap, titled).value(),
+                                () -> cap,
+                                () -> new TextOf(new StringBuilder()
+                                    .append(titled.asString())
+                                    .append(new Sub(text, 1))
+                                    .toString()
+                                    )
```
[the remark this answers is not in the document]
kokodyn: @victornoel done

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: @victornoel done

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: @andreoss While working on that PR I realized that there is a difference between Upper and TitleCase:
https://stackoverflow.com/questions/31770995/difference-between-uppercase-and-titlecase

victornoel: @kokodyn I agree with @andreoss, why not use the much simpler example he proposes?
- no need for testing if the first character is already title case or not, let's just do it as in @andreoss example above
- let's use Upper for now
- let's add a `@todo` to the class so that we add `Titled` and replace `Upper` in this class.

kokodyn: @victornoel @andreoss 
I will simplify logic as you suggested.
I created new PR to add Titled, could you look at it, please?
 I think firstly we can merge Titled and use it in Capitalized PR, ok?

victornoel: @kokodyn that's acceptable, but please take note that the purpose of cactoos is to be a sandbox area for experimenting with Zerocracy, on top of implementing this library. Hence in the future it would be best to follow the proper workflow which advocates for using `@todo`s as a way to decompose work. Once from time to time, it is ok to not follow this, but this should remain the general nominal way of working here.

kokodyn: @victornoel 
i added @TODO comment to with information what should be replaced in future.
I would like to have it proper working(and later replaced using static titled by "EO Titled" ) instead of using Lower.
As I said before this is tricky difference beetwen Lower and Titled.
For instance:
Title case: ǅ ->https://www.compart.com/en/unicode/U+01C5
Lower case: ǆ -> https://www.compart.com/en/unicode/U+01C6
Upper case: Ǆ -> https://www.compart.com/en/unicode/U+01C4

However, if you are sure that Lower should be used please let me know.

victornoel: @kokodyn I thought we were talking about `Upper` vs `Titled`, no? What I propose is to tackle this issue incrementally, so first we make it work with Upper because it's good enough, and then with the todo we explore how to have Capitalized use titled characters.

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,74 @@
...
+                () -> {
+                    return new Ternary<Text>(
+                        new IsBlank(text),
+                        () -> text,
+                        () -> new Joined(
+                            "",
+                            new TextOf(
+                                Character.toChars(
+                                    Character.toTitleCase(
+                                        new Sub(text, 0, 1).asString().codePointAt(0)
```
victornoel, line 65: @kokodyn I believe you can directly write `new TextOf(Character.toTitleCase(text.asString().codePointAt(0)))` no?

src/main/java/org/cactoos/text/Capitalized.java
victornoel: @kokodyn anywa, see my last comment about the current code, maybe we can just use title case as-is right now without introducing any new class.

kokodyn: Yes we can use tittle case without introducing any more class.

src/main/java/org/cactoos/text/Capitalized.java
kokodyn: Good point!
new TextOf(Character.toChars(Character.toTitleCase(text.asString().codePointAt(0))))
I believe this is the easiest version

victornoel:
@kokodyn I feel like we arrived to a good result :)

@andreoss could you take a last look to see if I haven't overlooked anything? thx

victornoel:
@andreoss could you finish the review?

src/main/java/org/cactoos/text/Capitalized.java
```diff
@@ -0,0 +1,72 @@
...
+                            "",
+                            new TextOf(
+                                Character.toChars(
+                                    Character.toTitleCase(
+                                        text.asString().codePointAt(0)
+                                    )
+                                )
+                            ),
+                            new Sub(text, 1)
+                            )
```
andreoss, line 68: @kokodyn The indentation looks odd here. And at 70 as well. 
You can also use lambda without `return` here.

andreoss (changes requested):

andreoss:
@victornoel Everything but the indentation looks good.

victornoel:
@andreoss thx, I will merge as-is since it is not critical and I will most certainly touch this same code when solving #1460 :)

victornoel:
@rultor merge

victornoel:
@kokodyn thx for your patience

0crat:
Code review was too long (15 days), architects (@victornoel) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

0crat:

sereshqua:
@kokodyn please make sure you start all your comments with the name of the user they are referred to, [see](https://github.com/yegor256/cactoos/pull/1465#discussion_r491097857)

sereshqua:
@0crat quality acceptable

kokodyn:
@victornoel @andreoss @sereshqua 
Thanks for participate, comments etc.
i will try to make better PR more and more.
BR,
kokodyn

Issue #1466, opened 2020-09-18
Add Capitalized EE version of StringUtils.capitalize


kokodyn:
Hi all
I would like to add  EO version of StringUtils.capitalize
[StringUtils.capitalize](http://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#capitalize-java.lang.String-)
But it's seems first I need to add class "Titled" to handle title case converting.
Is it ok for you?

I hope for your understanding, this is my first pull request to open source project.

victornoel:
@kokodyn I'm not exactly sure what you mean :)

I see there is #1465 which introduce `Capitalized`: is this PR outdated? It seems to implement `Capitalized` as you want it, no? Or do you mean that to simplify the PR, you would need `Titled`?

victornoel:
@kokodyn can you explain?

victornoel:
@kokodyn from what I understand, I believe we can handle all of that from the PR you made, see my comment https://github.com/yegor256/cactoos/pull/1465#discussion_r495467732. No need for a new ticket :)

kokodyn:
@victornoel 
To simplify logic I created new PR to introduce Titled.

victornoel:
@kokodyn I'm going to close this then. Please next time try to take things in the right order either by creating first a ticket for a new requirement not supported by cactoos or if you contribute a PR, by going iteratively and incrementally with `@todo`s.

We are in no rush to get PRs and solutions, we must first examine the problems as they come :)

By the way @kokodyn, woudln't you be interested in joining zerocracy and the cactoos project in order to contribute to it as part of the team? It would be great to have more DEV :)
See http://www.zerocracy.com/faq.html

PR #1474, opened 2020-09-27
 #1466 Added titled

kokodyn:
Converts the text into titled case.
Added Title which is required to my other PR(Added Capitalized).

### DIFF
--- a/src/main/java/org/cactoos/text/Titled.java
+++ b/src/main/java/org/cactoos/text/Titled.java
@@ -1,0 +1,59 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2020 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


victornoel:
@kokodyn I'm a bit worried about the naming because from a user point of view, if I read the term `Titled`, I'm expecting the string to be transformed to a title, i.e. with the first character of each word being titled case. Do you have other idea to name this?

IMHO it would have been best to first merge the other PR #1465 with a `@todo` and then we could have taken the time to think about this problem.... maybe we need another abstraction for characters that shouldn't be confused with `Text`. It's not clear.

victornoel:
@kokodyn I'm going to close this as #1465 is being merged. Also don't hesitate to join Zerocracy and this project if you want to contribute here and learn to use zerocracy :) see http://www.zerocracy.com/faq.html for details!

0crat:
Job `gh:yegor256/cactoos#1474` is not assigned, can't get performer

kokodyn:
@victornoel Thanks for invitation to "join Zerocracy" I will consider it :)

Issue #1475, opened 2020-09-27
Let's make Ternary able to have the same sticky value for each evaluation


victornoel:
Since most of the cactoos classes are lary-evaluated, sometimes when using `Ternary`, each call to each of the `Scalar` it holds in its state will result of a different evaluation of the input value.

For example, the meaning of the following is not very clear:
```java
final Scalar<Integer> scl = new ScalarOf<>(new AtomicInteger(0)::incrementAndGet);
new Ternary<String>(
    () -> scl.value() == 1,
    () -> scl.value() + " equals " + 1,
    () -> "else: " + scl.value()
).value() // returns "2 equals 1"
```

It is currently possible to write:
```java
final Scalar<Integer> scl = new ScalarOf<>(new AtomicInteger(0)::incrementAndGet);
new Ternary<String>(
    scl.value(),
    i -> i == 1,
    i -> i + " equals " + 1,
    () -> "else: " + i
).value() // returns "1 equals 1"
```

But it forces us to evaluate scl, which goes against the idea of only composing objects. This means we can't use it in constructors for example where we want to avoid code in them.

Instead, we would like to be able to write:
```java
final Scalar<Integer> scl = new ScalarOf<>(new AtomicInteger(0)::incrementAndGet);
new Ternary<String>(
    scl,
    i -> i == 1,
    i ->  i + " equals " + 1,
    i -> "else: " + i
).value() // return "1 equals 1"
```

With `i` being stable the whole `value()` evaluation.

It would be for example useful to tackle #1460.

victornoel:
@0crat assign me

victornoel:
@0crat wait for #1476

PR #1476, opened 2020-09-27
(#1475) Make Ternary consistent over its evaluation

victornoel:
This is for #1475.

I've also fixed a test that looked very fishy.

### DIFF
--- a/src/main/java/org/cactoos/scalar/Ternary.java
+++ b/src/main/java/org/cactoos/scalar/Ternary.java
@@ -25,6 +25,7 @@ package org.cactoos.scalar;
 
 import org.cactoos.Func;
 import org.cactoos.Scalar;
+import org.cactoos.func.FuncOf;
 
 /**
  * Ternary operation.
@@ -39,51 +40,17 @@ import org.cactoos.Scalar;
  *
  * <pre>{@code
  * new Ternary<>(
- *     5,
- *     input -> input > 3,
- *     input -> input = 8,
- *     input -> input = 2
- * ).value() // will be equal to 8
+ *     scalar,
+ *     value -> value > 3,
+ *     value -> true,
+ *     value -> false
+ * ).value() // will be equal to true
  * }</pre>
  *
  * @param <T> Type of item.
  * @since 0.8
  */
-public final class Ternary<T> implements Scalar<T> {
-
-    /**
-     * The condition.
-     */
-    private final Scalar<Boolean> condition;
-
-    /**
-     * The consequent.
-     */
-    private final Scalar<T> consequent;
-
-    /**
-     * The alternative.
-     */
-    private final Scalar<T> alternative;
-
-    /**
-     * Ctor.
-     * @param input The input to pass to all of them
-     * @param cnd The condition
-     * @param cons The consequent
-     * @param alter The alternative
-     * @param <X> Type of input
-     * @since 0.9
-     * @checkstyle ParameterNumberCheck (5 lines)
-     */
-    public <X> Ternary(final X input, final Func<X, Boolean> cnd,
-        final Func<X, T> cons, final Func<X, T> alter) {
-        this(
-            () -> cnd.apply(input),
-            () -> cons.apply(input),
-            () -> alter.apply(input)
-        );
-    }
+public final class Ternary<T> extends ScalarEnvelope<T> {
 
     /**
      * Ctor.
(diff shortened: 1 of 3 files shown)


src/test/java/org/cactoos/scalar/TernaryTest.java
```diff
@@ -73,14 +76,40 @@ void conditionBoolean() throws Exception {
...
+    @Test
+    void conditionSticky() throws Exception {
+        final AtomicInteger counter = new AtomicInteger(0);
+        new Assertion<>(
+            "Must conserve the same value for the whole evaluation",
+            new Ternary<String>(
+                counter::incrementAndGet,
+                i -> i == 1,
+                i ->  i + " equals " + 1,
+                i -> "else: " + i
```
andreoss, line 100: @victornoel It would be better as `String.format("else: %d")` here and in other pleases where `+` is used for concatenation.

src/test/java/org/cactoos/scalar/TernaryTest.java
```diff
@@ -73,14 +76,40 @@ void conditionBoolean() throws Exception {
...
+            ),
+            new ScalarHasValue<>("1 equals 1")
+        ).affirm();
+        new Assertion<>(
+            "Must conserve the same new value for the whole evaluation",
```
andreoss, line 105: @victornoel It's not clear here what should be conserved. Shouldn't `Ternary` be assigned to a variable for `Sticky` to have any effect?

src/main/java/org/cactoos/scalar/Ternary.java
```diff
@@ -39,10 +39,10 @@
  *
  * <pre>{@code
  * new Ternary<>(
- *     5,
+ *     someScalarOfInteger,
```
andreoss, line 42: @victornoel Camel-cased variables are against the current code style, probably we should not have them in examples too.

andreoss (changes requested):

src/test/java/org/cactoos/scalar/TernaryTest.java
victornoel: @andreoss ok, I see it's not clear. I thought the purpose of the PR was clear based on the ticket description, but maybe it's only clear in my head ^^ I will try to explain here and then once it's clear also for you, I will imrpove the code.

Basically, when you can `new Ternary(xxx).value()`, you want that for every branches of the ternary, the value used to be the same. So for example in the test, each call to the provided scalar.value will return a new value, but you want that each call to ternary.value gives you the result for one given value of the provided scalar.value. Like this it makes behaviour much closer to the expected one by someone discovering the class ternary.

Also it allows to express things that were not possible before, we will see that in the PR for #1460: it will allow to write it only by pure object composition without imperative intermediate values:
```java
     public Abbreviated(final Text text, final int max) {
         super(
             new Flatten(
                new Ternary<Text>(
                    () -> new Sticky(text),
                    t -> t.asString().length() <= max,
                    t -> t,
                    t -> new FormattedText(
                        "%s%s",
                        new Sub(
                            t,
                            0,
                            max - Abbreviated.ELLIPSES.length()
                        ),
                        Abbreviated.ELLIPSES
                    )
                )
             )
        )
    }
```

victornoel:
@andreoss sorry for the lateness, I will push some changes soon :)

src/test/java/org/cactoos/scalar/TernaryTest.java
victornoel: @andreoss actually, thank you for your comment because I realised afterward that it's not that it was clear, but as you said, that the test was incorrect: ternary should be assigned to a variable to be reused in the two assertions. And when I tried, it showed that my implementation was indeed incorrect :)

I've fixed that in my last push!

src/test/java/org/cactoos/scalar/TernaryTest.java
victornoel: @andreoss I replaced all of this with `FormattedText`

src/main/java/org/cactoos/scalar/Ternary.java
victornoel: @andreoss right, I also simplified the example

victornoel:
@andreoss please see my last push, all should be better now :)

andreoss (approved):
@victornoel

victornoel:
@andreoss thx for the review, it's always a pleasure to find bugs thanks to the reviewer :D

victornoel:
@rultor merge

0crat:

sereshqua:
@0crat quality good

Issue #1459, opened 2020-09-13
ListIteratorNoNulls.java:36-37: Add also some null checks...


0pdd:
The puzzle `1247-25d91c9c` from #1247 has to be resolved: 

https://github.com/yegor256/cactoos/blob/07b063e6af19ca733fd9b7afd38c8506f1c816f8/src/main/java/org/cactoos/list/ListIteratorNoNulls.java#L36-L37

The puzzle was created by Victor Noël on 13-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
This puzzle inherited milestone `1.0` from issue #1247.

0pdd:
The puzzle `1247-25d91c9c` has disappeared from the source code, that's why I closed this issue.

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @fabriciofx[/z](https://www.0crat.com/u/fabriciofx), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality good

PR #1478, opened 2020-10-02
#1459: ListIteratorNoNulls checks for #add() and #set()

fabriciofx:
Per #1459 :

- Add no null check for `#add(`) and `#set()`
- Add tests to assure they're working properly
- Remove the puzzle

### DIFF
--- a/src/main/java/org/cactoos/list/ListIteratorNoNulls.java
+++ b/src/main/java/org/cactoos/list/ListIteratorNoNulls.java
@@ -33,8 +33,6 @@ import java.util.ListIterator;
  *
  * @param <T> Element type
  * @since 0.39
- * @todo #1247:30min Add also some null checks both to the `set` and
- *  the `add` methods, then also add some tests to validate this behaviour.
  */
 public final class ListIteratorNoNulls<T> implements ListIterator<T> {
 
@@ -97,11 +95,21 @@ public final class ListIteratorNoNulls<T> implements ListIterator<T> {
 
     @Override
     public void set(final T item) {
+        if (item == null) {
+            throw new IllegalArgumentException(
+                "Item can't be NULL in #set(T)"
+            );
+        }
         this.listiterator.set(item);
     }
 
     @Override
     public void add(final T item) {
+        if (item == null) {
+            throw new IllegalArgumentException(
+                "Item can't be NULL in #add(T)"
+            );
+        }
         this.listiterator.add(item);
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/cactoos/list/ListIteratorNoNullsTest.java
```diff
@@ -135,4 +135,38 @@ void mustSetValueListIterator() {
             new ScalarHasValue<>(4)
         ).affirm();
     }
+
+    @Test
+    void mustThrowsErrorIfAddANullItem() {
+        new Assertion<>(
+            "must throw error if add a null item",
```
andreoss, line 142: @fabriciofx That's Exception not an Error

src/main/java/org/cactoos/list/ListIteratorNoNulls.java
```diff
@@ -97,11 +95,21 @@ public void remove() {
 
     @Override
     public void set(final T item) {
+        if (item == null) {
```
andreoss, line 98: @fabriciofx Using a static check method is not an option (i.e `Objects.requireNonNull`) here?
This check for `NULL` seems to be copy-pasted in many places.

andreoss (changes requested):

andreoss:
@fabriciofx Can you take a look at my comments?

victornoel:
@fabriciofx ping :)

src/main/java/org/cactoos/list/ListIteratorNoNulls.java
victornoel: @andreoss since this is the main responsibility of the class to check for null, I don't think it's really a problem. Also we tend to avoid using static methods such as `Objects.requireNonNull` when possible (I agree there are case it is just better to use the Java lib of course) and here it seems it is possible. In particular, it allows us to decide what is the error message here :)

victornoel:
@fabriciofx if we could finish this one too it would be great :)

victornoel:
@fabriciofx ping :)

victornoel:
@fabriciofx ping :)

victornoel:
@andreoss I think I will merge this as-is if you are ok. It seems the only real problem here is a not perfect test message, it seems acceptable to leave it like that IMHO.

victornoel:
@rultor merge

0crat:
Code review was too long (87 days), architects (@victornoel) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

0crat:

sereshqua:
@andreoss please make sure you will find at least 3 issues during next CR, thanks

sereshqua:
@0crat quality acceptable

PR #1484, opened 2020-10-15
Add IsEmpty to check if Text is empty or not

kokodyn:
I would like to introduce a class in order to check whether the text is empty or not

### DIFF
--- a/src/main/java/org/cactoos/text/IsEmpty.java
+++ b/src/main/java/org/cactoos/text/IsEmpty.java
@@ -27,7 +27,8 @@ import org.cactoos.Scalar;
 import org.cactoos.Text;
 
 /**
- * Determines if text is empty or no.
+ * Determines if text is empty or not.
+ * @see IsBlank
  * @since 0.47
  */
 public final class IsEmpty implements Scalar<Boolean> {
--- a/src/main/java/org/cactoos/text/IsEmpty.java
+++ b/src/main/java/org/cactoos/text/IsEmpty.java
@@ -1,0 +1,52 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2020 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.text;
+
+import org.cactoos.Scalar;
+import org.cactoos.Text;
+
+/**
+ * Determines if text is empty or no.
+ * @since 0.47
+ */
+public final class IsEmpty implements Scalar<Boolean> {
+
+    /**
+     * The text.
+     */
+    private final Text origin;
+
+    /**
+     * Ctor.
+     * @param text The text
+     */
+    public IsEmpty(final Text text) {
+        this.origin = text;
+    }
+
+    @Override
+    public Boolean value() throws Exception {
+        return this.origin.asString().isEmpty();
+    }
+}
--- a/src/main/java/org/cactoos/text/IsEmpty.java
+++ b/src/main/java/org/cactoos/text/IsEmpty.java
@@ -35,18 +35,18 @@ public final class IsEmpty implements Scalar<Boolean> {
     /**
      * The text.
      */
-    private final Text origin;
+    private final Text txt;
 
     /**
      * Ctor.
      * @param text The text
      */
     public IsEmpty(final Text text) {
-        this.origin = text;
+        this.txt = text;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/cactoos/text/IsEmpty.java
```diff
@@ -0,0 +1,52 @@
...
+package org.cactoos.text;
+
+import org.cactoos.Scalar;
+import org.cactoos.Text;
+
+/**
+ * Determines if text is empty or no.
+ * @since 0.47
+ */
+public final class IsEmpty implements Scalar<Boolean> {
```
andreoss, line 33: @kokodyn 
How is this class practical? Can we use it somewhere in the code base?

For example we could use `Func<Text, Boolean>` here https://github.com/llorllale/cactoos-matchers/blob/master/src/main/java/org/llorllale/cactoos/matchers/HasValuesMatching.java#L56. I guess technically you can pass it there as a method reference `IsEmpty::new`. 
But having `Scalar<Boolean>` in order to check is a bad idea imo, it creates short-lived objects.

andreoss (changes requested):

src/test/java/org/cactoos/text/IsEmptyTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "Must't detect nonempty text",
+            new IsEmpty(new TextOf("abc")),
+            new ScalarHasValue<>(Boolean.FALSE)
+        ).affirm();
+    }
+
+    @Test
+    void determinesBlankText() {
+        new Assertion<>(
+            "Must't detect blank text",
```
andreoss, line 57: @kokodyn Some kind of typo here and at 48

src/main/java/org/cactoos/text/IsEmpty.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * The text.
+     */
+    private final Text origin;
```
andreoss, line 38: @kokodyn `origin` is a good name for a decorated object. Here it's not the case

andreoss (changes requested):

src/main/java/org/cactoos/text/IsEmpty.java
victornoel: @kokodyn @andreoss let's keep it as a scalar for now, if short lived objects were a problem we would have to throw most of cactoos away :) (btw the JVM is very good at optimizing memory and execution for short lived objects)

src/main/java/org/cactoos/text/IsEmpty.java
kokodyn: @andreoss 
I see in most classes implementing Scalar we use "orgin". For example 
https://github.com/yegor256/cactoos/blob/master/src/main/java/org/cactoos/text/IsBlank.java
I have no idea for a better name
Could you propose better name, please?

src/main/java/org/cactoos/text/IsEmpty.java
kokodyn: @andreoss @victornoel 
Practical using IsEmpty class to create matcher in org.llorllale.cactoos.matchers;
IsEmpty()
and then instead of using of new matcher:
        new Assertion<>(
            "The result text wasn't empty",
            new TextOf(input),
            new TextIs("")
        ).affirm();`
Like here:
https://github.com/yegor256/cactoos/blob/master/src/test/java/org/cactoos/io/HeadInputStreamTest.java#L79
We would use:
`        new Assertion<>(
            "The result text wasn't empty",
            new TextOf(input),
            new IsEmpty()
        ).affirm();`
I think it is much more readable with using IsEmpty

src/main/java/org/cactoos/text/IsEmpty.java
andreoss: @kokodyn I would use `txt` or `text`. `origin` seems strange for a regular dependency.

kokodyn: @andreoss fixed

kokodyn:
Hi @andreoss @victornoel 
I see some checks were not successful. I do not know what should I proceed in such a situation. My changes seems not related to: src/main/java/org/cactoos/scalar/Solid.java
So I do not understand why it is marked as impacted file and coverge is mesured.
Could you please suggest something, what are the next steps to improve it?
--

victornoel:
@kokodyn don't worry about it, it's a recurring situation. I will try to fix it for future pull requests in cactoos.

kokodyn:
@victornoel Thanks for quick answer.
For me it looks ok and if possible it would be merged.

victornoel:
@kokodyn let's wait for REV validation :)

andreoss (approved):
@victornoel 
Looks okay now.

src/main/java/org/cactoos/text/IsEmpty.java
```diff
@@ -0,0 +1,52 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.text;
+
+import org.cactoos.Scalar;
+import org.cactoos.Text;
+
+/**
+ * Determines if text is empty or no.
```
victornoel, line 30: @kokodyn proper english is "empty or not"

src/main/java/org/cactoos/text/IsEmpty.java
```diff
@@ -0,0 +1,52 @@
...
+ * @since 0.47
```
victornoel, line 31: @kokodyn can you add a `@see` for `IsBlank` (and one in `IsBlank` for `IsEmpty`).

src/test/java/org/cactoos/text/IsEmptyTest.java
```diff
@@ -0,0 +1,62 @@
...
+            "Must determine empty text",
+            new IsEmpty(new TextOf("")),
+            new ScalarHasValue<>(Boolean.TRUE)
+        ).affirm();
+    }
+
+    @Test
+    void determinesNonemptyText() {
+        new Assertion<>(
+            "Mustn't detect nonempty text",
```
victornoel, line 48: @kokodyn can you not abbreviate and just put "Must not" (same in the test below) and also "non-empty".

victornoel (changes requested):
@kokodyn a few more details :) thanks!

src/test/java/org/cactoos/text/IsEmptyTest.java
kokodyn: @victornoel done

src/main/java/org/cactoos/text/IsEmpty.java
kokodyn: @victornoel done

src/main/java/org/cactoos/text/IsEmpty.java
kokodyn: @victornoel done

victornoel:
@kokodyn thanks :)

victornoel:
@rultor merge

0crat:

sereshqua:
@0crat quality good

PR #1480, opened 2020-10-06
Added Concatenated class to allow concatenate texts

kokodyn:
Hi
I would like to propose to add new class which allow concatenate texts.
Signed-off-by: kokodyn <kokodyn@gmail.com>

Issue #1051, opened 2019-01-29
Concatenated as an alternative to StringBuilder


Iprogrammerr:
Currently, we do not have Text implementation that works like StringBuilder. There is a Joined, but he concatenates strings/texts with a given delimit. I propose to create Concatenated that will work like this:
`new Concatenated(
    "SELECT * FROM A ",
    "WHERE id = 1"
).asString()`
He should accept both Strings and Texts.

0crat:
@llorllale[/z](https://www.0crat.com/u/llorllale) please, pay attention to this issue

fabriciofx:
@Iprogrammerr What's the problem with `JoinedText`? It does exactly the same thing.

Iprogrammerr:
@fabriciofx No, it does not. To have concatenated without delimit strings, I have to do
`new Joined("", "a",b","c)` which will result in `"a" + "" + "b" + "" + "c"`. I am suggesting replacement for StringBuilder, so that  `new Concatenated("a", "b", "c")` will result in `"a" + "b" + "c"`.

fabriciofx:
@Iprogrammerr I see now. Thanks!

llorllale:
@Iprogrammerr is this extra class really worth the trouble of not having to specify a delimiter for `org.cactoos.text.Joined`? I'm having trouble believing it.

llorllale:
@Iprogrammerr ping

How about we use the empty string as a default delimiter?

kokodyn:
Hi
I see i has been partly covered in PR #1480. Adding ctor which accepts String vargs in ctor it is what need to be implement.

victornoel:
@Iprogrammerr it seems to be implemented now

0crat:
Job `gh:yegor256/cactoos#1051` is not assigned, can't get performer

PR #1511, opened 2020-12-05
#1051 add ctor with vargs String

kokodyn:
Change related with #1051. I see i has been partly covered in PR #1480. Adding ctor which accepts String vargs in ctor is also required.

### DIFF
--- a/src/main/java/org/cactoos/text/Concatenated.java
+++ b/src/main/java/org/cactoos/text/Concatenated.java
@@ -25,6 +25,7 @@ package org.cactoos.text;
 
 import org.cactoos.Text;
 import org.cactoos.iterable.IterableOf;
+import org.cactoos.iterable.Mapped;
 
 /**
  * Concatenate a Text.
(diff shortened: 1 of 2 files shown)


src/main/java/org/cactoos/text/Concatenated.java
```diff
@@ -43,6 +44,19 @@ public Concatenated(final Text... txts) {
...
 
+    /**
+     * Ctor.
+     * @param strs Strings to be concatenated
+     */
+    public Concatenated(final String... strs) {
+        this(
+            new Mapped<>(
+                TextOf::new,
+                new IterableOf<String>(strs)
```
andreoss, line 55: @kokodyn Do you need explicit type parameter here?

andreoss (changes requested):

src/main/java/org/cactoos/text/Concatenated.java
kokodyn: @andreoss fixed

victornoel:
@rultor merge

victornoel:
@kokodyn @andreoss thanks

Issue #1464, opened 2020-09-15
Paged.java:44-46: Continue refactoring and add...


0pdd:
The puzzle `1183-64533d21` from #1183 has to be resolved: 

https://github.com/yegor256/cactoos/blob/af979def342f0bfe75e31912588681cd090c8c6c/src/main/java/org/cactoos/iterable/Paged.java#L44-L46

The puzzle was created by andreoss on 09-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0crat:
The job [#1464](https://github.com/yegor256/cactoos/issues/1464) assigned to @baudoliver7[/z](https://www.0crat.com/u/baudoliver7), here is [why](https://www.0crat.com/footprint/C63314D6Z/a7398e5c-ec73-4e8d-96cf-9a3e56933bb6); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

0pdd:
The puzzle `1183-64533d21` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd 2 puzzles [#1537](https://github.com/yegor256/cactoos/issues/1537), [#1538](https://github.com/yegor256/cactoos/issues/1538) are still not solved.

0crat:
@sereshqua[/z](https://www.0crat.com/u/sereshqua) please review this job completed by @baudoliver7[/z](https://www.0crat.com/u/baudoliver7), as in [§30](http://www.zerocracy.com/policy.html#30); the job will be fully closed and all payments will be made when the quality review is completed

sereshqua:
@0crat quality good

0pdd:
[...]

PR #1536, opened 2021-01-19
#1464 Add iterator.Paged and remove checkstyle suppression

baudoliver7:
#1464 

- create `iterator.Paged`
- refactor `iterable.Paged`
- remove checkstyle suppression in `iterable.Paged`

### DIFF
--- a/src/main/java/org/cactoos/iterator/Paged.java
+++ b/src/main/java/org/cactoos/iterator/Paged.java
@@ -1,0 +1,87 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2020 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.iterator;
+
+import java.util.Iterator;
+import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicReference;
+import org.cactoos.Func;
+import org.cactoos.func.UncheckedFunc;
+
+/**
+ * Paged iterator.
+ * Elements will continue to be provided so long as {@code next} produces
+ * non-empty iterators.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @param <I> Subtype of item iterator
+ * @param <X> Type of item
+ * @since 0.49
+ * @todo #1464:30min We want to use {@code Iterator<X>} directly in the class without
+ *  introducing I, maybe using {@code Iterator<? extends X>} for example because, we don't
+ *  have to constrain the {@link Func} to give exactly the same Iterator type as long as it
+ *  is an iterator that provides X. {@link org.cactoos.iterable.Paged} should be changed too.
+ */
+public final class Paged<I extends Iterator<X>, X> implements Iterator<X> {
+
+    /**
+     * Current element.
+     */
+    private final AtomicReference<I> current;
+
+    /**
+     * Function to get the next element.
+     */
+    private final Func<I, I> subsequent;
+
+    /**
+     * Ctor.
+     * @param first First element.
+     * @param next Function to get the next element.
+     */
+    public Paged(final I first, final Func<I, I> next) {
+        this.current = new AtomicReference<>(first);
+        this.subsequent = next;
+    }
+
+    @Override
+    public boolean hasNext() {
+        if (!this.current.get().hasNext()) {
+            final I next = new UncheckedFunc<>(this.subsequent).apply(
+                this.current.get()
+            );
+            this.current.set(next);
+        }
+        return this.current.get().hasNext();
+    }
+
+    @Override
+    public X next() {
+        if (this.hasNext()) {
+            return this.current.get().next();
+        }
+        throw new NoSuchElementException();
+    }
+}
--- a/src/main/java/org/cactoos/iterable/Paged.java
+++ b/src/main/java/org/cactoos/iterable/Paged.java
@@ -24,13 +24,7 @@
 package org.cactoos.iterable;
 
 import java.util.Iterator;
-import java.util.NoSuchElementException;
-import java.util.concurrent.atomic.AtomicReference;
 import org.cactoos.Func;
-import org.cactoos.Scalar;
-import org.cactoos.func.UncheckedFunc;
-import org.cactoos.scalar.Sticky;
-import org.cactoos.scalar.Unchecked;
 
 /**
  * Paged iterable.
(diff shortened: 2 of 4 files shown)


0crat:
@victornoel[/z](https://www.0crat.com/u/victornoel) everybody who has role `REV` is banned at [#1536](https://github.com/yegor256/cactoos/issues/1536); I won't be able to assign anyone automatically; consider assigning someone manually (as in [§19](http://www.zerocracy.com/policy.html#19)), or invite more people (as in [§51](http://www.zerocracy.com/policy.html#51)), or remove the job from the scope (as in [§14](http://www.zerocracy.com/policy.html#14))

baudoliver7:
@victornoel PR done

src/main/java/org/cactoos/iterator/Paged.java
```diff
@@ -0,0 +1,95 @@
...
+ * Elements will continue to be provided so long as {@code next} produces
+ * non-empty iterators.
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @param <I> Subtype of item iterator
+ * @param <X> Type of item
+ * @since 0.49
+ */
+public final class Paged<I extends Iterator<X>, X> implements Iterator<X> {
```
victornoel, line 46: @baudoliver7 why do we need this `I` here, can't we just use `Iterator<X>` directly in the class without introducing `I`? If it's doable, let's get rid of it :) Or instead just add a todo for creating a new task for it. I think we should try to get rid of it here and also in iterable.Paged, maybe using `Iterator<? extends X>` for example, because we don't have to constrain the `Func` to give exactly the same Iterator type as long as it  is an iterator that provides `X`.

src/main/java/org/cactoos/iterator/Paged.java
```diff
@@ -0,0 +1,95 @@
...
+
+    /**
+     * Current element.
+     */
+    private final AtomicReference<Unchecked<I>> current;
+
+    /**
+     * Function to get the next element.
+     */
+    private final UncheckedFunc<I, I> subsequent;
```
victornoel, line 56: @baudoliver7 just for clarity, I believe it's best to store here the `Func` directly, and wrap in unchecked only in `hasNext`

src/main/java/org/cactoos/iterator/Paged.java
```diff
@@ -0,0 +1,95 @@
...
+ * @param <I> Subtype of item iterator
+ * @param <X> Type of item
+ * @since 0.49
+ */
+public final class Paged<I extends Iterator<X>, X> implements Iterator<X> {
+
+    /**
+     * Current element.
+     */
+    private final AtomicReference<Unchecked<I>> current;
```
victornoel, line 51: @baudoliver7 just for clarity, I believe it's best to store here the `Scalar` directly, and wrap in unchecked only in `hasNext` and `next`

src/main/java/org/cactoos/iterator/Paged.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public boolean hasNext() {
+        if (!this.current.get().value().hasNext()) {
+            final I next = this.subsequent.apply(
+                this.current.get().value()
+            );
+            this.current.set(
+                new Unchecked<>(
+                    new Sticky<>(() -> next)
+                )
```
victornoel, line 82: @baudoliver7 it seems to me that we don't need at all to wrap into a `Scalar`, let's just store `next` itself here

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel I agree with you. I do it now.

baudoliver7: @victornoel After having thought, I found out it is not doable to get rid of it unless we want to replace `I` everywhere it is used by `? extends Iterator<X>`, which is not easy to read. It's for the sake of clarity that `I` has been introduced.

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel I see but the reason why it's done like this is for the need of storing the current value. By wanting to respect the principle of immutability, we need to store somewhere the current item after calling `hasNext`. And AtomicReference coupled to Unchecked serve the purpose of it.

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel `next` is just the function that retrieve the next item. The value found is stored in `current` variable that is a `AtomicReference`. We wrap the `Scalar` like this to give us possibility to update its value later while keeping it immutable.

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel Good :)

baudoliver7:
@victornoel Some changes have been made

src/main/java/org/cactoos/iterator/Paged.java
victornoel: @baudoliver7 the problem is that it is incorrect like this because it means we cannot produce different subclasses of iterator with this, only one that matches the `I`. Clarity is not important here, what is important is that the class is usable.

I propose the following: let's add a todo to get rid of this `I` so that we can now focus on the rest of the PR :)

src/main/java/org/cactoos/iterator/Paged.java
victornoel: @baudoliver7 you can hide mutability behind a scalar, but it is still mutability. We have a tendency to use `AtomicReference` and co in cactoos to avoid non-final fields, but then, if it's mutable, it's mutable, that's life.

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel Ok

src/main/java/org/cactoos/iterator/Paged.java
victornoel: @baudoliver7 I think this discussion is moot because of https://github.com/yegor256/cactoos/pull/1536#discussion_r560400853, so I resolve this discussion :)

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel Done.

src/main/java/org/cactoos/iterator/Paged.java
baudoliver7: @victornoel Yeah :)

src/main/java/org/cactoos/iterator/Paged.java
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @param <I> Subtype of item iterator
+ * @param <X> Type of item
+ * @since 0.49
+ *  @todo #1464:30min We want to use {@code Iterator<X>} directly in the class without
+ *  introducing I, maybe using {@code Iterator<? extends X>} for example because, we don't
+ *  have to constrain the {@link Func} to give exactly the same Iterator type as long as it
+ *  is an iterator that provides X .
```
victornoel, line 48: @baudoliver7 let's also add that `iterable.Paged` should be changed too

baudoliver7: @victornoel Done.

src/main/java/org/cactoos/iterator/Paged.java
victornoel: @baudoliver7 we are still missing that change to have everything for merging :)

baudoliver7: @victornoel In this case, `this.current` should be `non final`. Isn't it ?

victornoel: @baudoliver7 it should be:
```java
private final AtomicReference<I> current;
```

baudoliver7: @victornoel By this way, constructor will be : 
```
public Paged(final I first, final Func<I, I> next) {
    this.current = new AtomicReference<I>(first);
    this.subsequent = next;
}
```
Isn't it ?

victornoel: @baudoliver7 ha,  hmmm, looks like I didn't think this through... but then yes! I always thought it was strange to have a `Scalar` as first... an `Iterator` is already an object with behaviour by itself, no need for `Scalar` which is useful when you cannot have your own implementation of types such as `String` or `Integer`. You can of course also add a todo if you are sick of all those extra changes I am adding, this is what they are for :)

And now I really wonder why `iterable.Paged` is taking `Iterator` and not `Iterable`... ! Could you also add a todo in `iterable.Paged`: "Replace the constructor so that instead it takes a `Iterable` as `first` and a `Func<Iterable, Iterable>` as `next` in order to homogenise it with the other `Iterable` implementations." Thx!

baudoliver7: @victornoel I done all extra changes about using `I` directly and I reported `Replace the constructor...` as a todo.

0crat:
@victornoel[/z](https://www.0crat.com/u/victornoel) everybody who has role `REV` is banned at [#1536](https://github.com/yegor256/cactoos/issues/1536); I won't be able to assign anyone automatically; consider assigning someone manually (as in [§19](http://www.zerocracy.com/policy.html#19)), or invite more people (as in [§51](http://www.zerocracy.com/policy.html#51)), or remove the job from the scope (as in [§14](http://www.zerocracy.com/policy.html#14))

baudoliver7:
@victornoel Some changes have been made.

victornoel:
@baudoliver7 great, thx

victornoel:
@rultor merge

0crat:
Job `gh:yegor256/cactoos#1536` is not assigned, can't get performer

Issue #1461, opened 2020-09-13
TextOf.java:63-67: Introduce `Text` implementations...


0pdd:
The puzzle `1287-48d7a962` from #1287 has to be resolved: 

https://github.com/yegor256/cactoos/blob/47aff6e4c70f6f53d24bdccb7c1f6c12abfe60d2/src/main/java/org/cactoos/text/TextOf.java#L63-L67

The puzzle was created by Victor Noël on 13-Sep-20. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
This puzzle inherited milestone `1.0` from issue #1287.

0crat:
The job [#1461](https://github.com/yegor256/cactoos/issues/1461) assigned to @ruca905[/z](https://www.0crat.com/u/ruca905), here is [why](https://www.0crat.com/footprint/C63314D6Z/13f1d1d1-4d54-4c57-b002-dd5fd7c730f9); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

victornoel:
@0crat refuse

0crat:
> @0crat refuse ([here](https://github.com/yegor256/cactoos/issues/1461#issuecomment-778824319))

@victornoel The user @ruca905[/z](https://www.0crat.com/u/ruca905) resigned from [#1461](https://github.com/yegor256/cactoos/issues/1461), please stop working. Reason for job resignation: Order was cancelled

0pdd:
The puzzle `1287-48d7a962` has disappeared from the source code, that's why I closed this issue.

0crat:
[...]

PR #1547, opened 2021-02-25
(#1461) Introduce Text implementations

baudoliver7:
#1461

- Implement `TextOfScalar`, `TextOfString`, `TextOfBytes`, `TextOfDate`, `TextOfIterable` and `TextOfIterator`
- Refactor `TextOf` to use those new classes


0crat:
@victornoel[/z](https://www.0crat.com/u/victornoel) everybody who has role `REV` is banned at [#1547](https://github.com/yegor256/cactoos/issues/1547); I won't be able to assign anyone automatically; consider assigning someone manually (as in [§19](http://www.zerocracy.com/policy.html#19)), or invite more people (as in [§51](http://www.zerocracy.com/policy.html#51)), or remove the job from the scope (as in [§14](http://www.zerocracy.com/policy.html#14))

src/main/java/org/cactoos/text/TextOfDate.java
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * @since 0.49
+ */
+public final class TextOfDate extends TextEnvelope {
+
+    /**
+     * Formats the date using the provided formatter.
+     * @param date The date to format.
+     * @param formatter The formatter to use.
+     * @since 1.0
```
victornoel, line 48: @baudoliver7 let's just use `@since 1.0.0` on every of those classes and not on the constructors at all. We will in the future fix it for for the rest of the codebase

src/main/java/org/cactoos/text/TextOfIterable.java
```diff
@@ -0,0 +1,55 @@
...
+    public TextOfIterable(final Iterable<?> iterable) {
+        super(
+            new TextOfScalar(
+                () -> new Joined(
+                    ", ",
+                    new Mapped<>(
+                        Object::toString,
+                        iterable
+                    )
+                ).asString()
```
victornoel, line 51: @baudoliver7 here the idea is that you can directly pass the Joined object to super without using `TextOfScalar`. that's why there was this todo: to avoid using scalars when not needed.

src/main/java/org/cactoos/text/TextOfString.java
```diff
@@ -0,0 +1,46 @@
...
+public final class TextOfString extends TextEnvelope {
+
+    /**
+     * Ctor.
+     *
+     * @param input The String
+     * @param cset The Charset
+     */
+    public TextOfString(final String input, final Charset cset) {
+        super(new TextOfScalar(() -> new String(input.getBytes(cset), cset)));
```
victornoel, line 44: @baudoliver7 this is quite strange, isn't this equivalent to just use the String itself? I would propose to rewrite this class by directly implementing `Text` instead of delegating to `TextOfScalar` and to remove the `Charset` which makes no sense to me.

src/main/java/org/cactoos/text/TextOfIterator.java
```diff
@@ -0,0 +1,53 @@
...
+     * @param iterator The iterable to convert to string
+     */
+    public TextOfIterator(final Iterator<Character> iterator) {
+        super(
+            new TextOfScalar(
+                () -> {
+                    final StringBuilder buf = new StringBuilder();
+                    iterator.forEachRemaining(buf::append);
+                    return buf.toString();
+                }
```
victornoel, line 49: @baudoliver7 this should instead delegate to `new TextOfIterable(new IterableOf(iterator))`, also I think we can just take an `Iterator<?>`

src/main/java/org/cactoos/text/TextOfBytes.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Ctor.
+     *
+     * @param bytes The Bytes
+     * @param cset The Charset
+     */
+    public TextOfBytes(final Bytes bytes, final String cset) {
+        super(
+            new TextOfScalar(
+                () -> new String(bytes.asBytes(), cset)
```
victornoel, line 61: @baudoliver7 this should be a secondary constructor delegating to the other one I think if possible

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -427,11 +403,7 @@ public TextOf(
      */
     public TextOf(final LocalDate date, final DateTimeFormatter formatter) {
         this(
-            () -> formatter.format(
-                ZonedDateTime.of(
-                    date, LocalTime.MIN, ZoneId.systemDefault()
-                )
-            )
+            new TextOfDate(date, formatter)
```
victornoel, line 406: @baudoliver7 it seems the `TextOfDate` class is specific enough for justifying not having those constructors in `TextOf` so that users directly uses `TextOfDate`

src/main/java/org/cactoos/text/TextOfDate.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.Date;
+
+/**
+ * Text of {@link Date}
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @since 0.49
+ */
+public final class TextOfDate extends TextEnvelope {
```
victornoel, line 42: @baudoliver7 let's class this class `TextOfDateTime` and have all constructors take `DateTimeFormatter` as the first argument.

victornoel (changes requested):
@baudoliver7 here are some comments and a more general one: the (I think not clear enough) objective for the todo was that by extracting some of the constructors as class, we could avoid using scalars in some case (e.g., the one with `Joined` for `Iterable`).

So I think it would be best to keep `TextOfIterable` and `TextOfIterator` inlined in `TextOf` instead of in their own class. It seems that actually this is also true for `TextOfBytes`, so let's move it back there :)

baudoliver7:
@victornoel Please, see my changes.

src/main/java/org/cactoos/text/TextOfString.java
```diff
@@ -0,0 +1,55 @@
...
+import org.cactoos.Text;
+
+/**
+ * Text of {@link String}
+ *
+ * <p>There is no thread-safety guarantee.
+ *
+ * @since 1.0.0
+ */
+public final class TextOfString implements Text {
```
victornoel, line 35: @baudoliver7 this class also needs equals, hashCode and toString implement

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -335,17 +315,7 @@ public TextOf(final String input) {
...
-        this(() -> new String(input.getBytes(cset), cset));
-    }
-
-    /**
-     * Ctor.
      * @param iterable The iterable to convert to string
-     * @since 0.21
      */
     public TextOf(final Iterable<?> iterable) {
         this(
```
victornoel, line 321: @baudoliver7 for this one, as I was proposing in my previous comment, let's directly delegate to Joined without using a lambda, and let's have `TextOf(final Iterator<?> iterator)` delegate to the `Iterable` constructor directly

src/main/java/org/cactoos/text/TextOf.java
```diff
@@ -384,276 +354,13 @@ public TextOf(final InputStream input) {
...
      * @param scalar The Scalar of String
      */
     public TextOf(final Scalar<String> scalar) {
-        this.origin = scalar;
-    }
-
-    @Override
-    public String asString() throws Exception {
-        return this.origin.value();
+        super(new TextOfScalar(scalar));
```
victornoel, line 363: @baudoliver7 the private constructor taking a `Text` was a good idea IMHO, instead of forcing other constructor to go through the use of a `Scalar` (for example the constructor for `Iterable`).

src/test/java/org/cactoos/text/TextOfDateTimeTest.java
```diff
@@ -0,0 +1,292 @@
...
+ * Test case for {@link TextOfDateTime}.
+ *
+ * @since 0.12
+ * @checkstyle JavadocMethodCheck (1000 lines)
+ * @checkstyle ClassDataAbstractionCouplingCheck (1000 lines)
+ * @checkstyle MagicNumberCheck (1000 lines)
+ * @checkstyle StringLiteralsConcatenationCheck (1000 lines)
+ */
+@SuppressWarnings({"PMD.TooManyMethods", "PMD.AvoidDuplicateLiterals"})
+final class TextOfDateTimeTest {
```
victornoel, line 55: @baudoliver7 let's rename the Can't into Must and also I think `JavadocMethodCheck` can be removed above

src/test/java/org/cactoos/text/TextOfDateTimeTest.java
```diff
@@ -0,0 +1,292 @@
...
+import org.hamcrest.core.IsNull;
+import org.junit.jupiter.api.Test;
+import org.llorllale.cactoos.matchers.Assertion;
+import org.llorllale.cactoos.matchers.IsBlank;
+import org.llorllale.cactoos.matchers.IsText;
+
+/**
+ * Test case for {@link TextOfDateTime}.
+ *
+ * @since 0.12
```
victornoel, line 48: @baudoliver7 let's use `@since 1.0.0` here and everywhere you introduced classes too

victornoel (changes requested):
@baudoliver7 some more comments in line with the previous ones :)

src/main/java/org/cactoos/text/TextOf.java
baudoliver7: @victornoel When we delegate `TextOf(final Iterator<?> iterator)` to the `Iterable` constructor directly, it changes code behaviour and fails this test :
```java
@Test
void readsIteratorToText() throws Exception {
    new Assertion<>(
        "Can't read Iterator to Text",
        new TextOf(
            new IteratorOfChars("qwer")
        ).asString(),
        new IsEqual<>("qwer")
    ).affirm();
}
```
with this error :
```cmd
[ERROR]   TextOfTest.readsIteratorToText:329 Can't read Iterator to Text
Expected: "qwer"
 but was: was "q, w, e, r"
```
FYI, that is what I changed in the first constructor :
```java
/**
 * Ctor.
 *
 * @param iterable The iterable to convert to string
 */
public TextOf(final Iterable<?> iterable) {
    super(
        new Joined(
            ", ",
            new Mapped<>(
                Object::toString,
                iterable
            )
        )
    );
}
```

victornoel: @baudoliver7 I think we should uniformise the behaviour of `TextOf` for `Iterator` and `Iterable`... that is something I overlooked when making previous comments :)
The best would be to add a todo and leave things as they are. But the todo should propose and idea of solution or at least be clear that it should be consistent with the fact that we already have `Concatenated` and `Joined`, so maybe we should only support `Iterable` and `Iterator` of `Character`?

src/main/java/org/cactoos/text/TextOf.java
baudoliver7: @victornoel I agree with you. But by bringing back `Bytes` constructors, there is a conflict between `Scalar` and `Text` constructors.

victornoel: @baudoliver7 we can use `TextOfScalar` directly to define the `Bytes` constructor I think

baudoliver7: @victornoel Right !

src/main/java/org/cactoos/text/TextOf.java
baudoliver7: @victornoel I agree to support only `Iterable` and `Iterator` of `Character` ?

victornoel: @baudoliver7 indeed, that seems to be the best!

baudoliver7: @victornoel It's already supported in my last commit :)

baudoliver7:
@victornoel Could see my answers and changes ?

src/main/java/org/cactoos/text/TextOf.java
victornoel: @baudoliver7 I meant to have support for `Iterable<Character>` and `Iterator<Character>`, both behaving similarly, most certainly the iterator one delegating to the iterable one

baudoliver7: @victornoel Use constructor for supporting `Iterable<Character>` fails most part of code. In fact, Constructor `Iterator<Character>` is use to form a word with a list of characters. However, Constructor `Iterable<?>` concat a list of string. Both behaviour are different.

victornoel: @baudoliver7 yes, the point of the proposed change is to modify actual behaviour of `TextOf` to simplify it. So it means that we need to modify other parts of cactoos that relied on this in order to preserve their behaviour (using joined our concatenated). Let's add a new todo for now to do it, ok?

baudoliver7: @victornoel Ok. I added a todo.

baudoliver7:
@victornoel Please, check my changes.

victornoel:
@baudoliver7 thx for your patience :)

victornoel:
@rultor merge

baudoliver7:
@victornoel Happy to contribute and learn :)

victornoel:
@rultor merge

0crat:

sereshqua:
@0crat quality bad

0crat:
Quality is low, no payment, see [§31](http://www.zerocracy.com/policy.html#31)

Issue #1507, opened 2020-11-16
MappedWithIndex would be nice to have


yegor256:
The functionality of `Mapped` is not enough, since it doesn't provide the index of the element into the `Func`. I'm suggesting to create `MappedWithIndex(BiFunc, items)`.

victornoel:
@yegor256 good idea, let's add it to iterator and iterable (delegating to the iterator one) then.

victornoel:
@0crat in

0crat:
The job [#1507](https://github.com/yegor256/cactoos/issues/1507) assigned to @dearrudam[/z](https://www.0crat.com/u/dearrudam), here is [why](https://www.0crat.com/footprint/C63314D6Z/fbe91dbc-a674-4836-85f3-7badd45fc778); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

dearrudam:
@yegor256 could you close this issue because it was solved with #1554?

yegor256:
@dearrudam thanks!

sereshqua:
@0crat quality good

PR #1554, opened 2021-03-04
For Issue #1507

dearrudam:
It's for issue #1507


src/main/java/org/cactoos/scalar/ScalarWithIndex.java
```diff
@@ -0,0 +1,68 @@
...
+import org.cactoos.Scalar;
+
+/**
+ * Element with its index.
+ *
+ * <p>There is no thread-safety guarantee.
+ * @param <T> Scalar type
+ * @since 0.50
+ */
+public final class ScalarWithIndex<T> implements Scalar<T> {
```
andreoss, line 35: @dearrudam Let's remove this, and use `MapEntry` instead

src/main/java/org/cactoos/iterator/MappedWithIndex.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * Mapped with index iterator.
+ *
+ * <p>
+ * There is no thread-safety guarantee.
+ * @param <Y> Type of target item
+ * @since 0.50
+ */
+public final class MappedWithIndex<Y> extends IteratorEnvelope<Y> {
```
andreoss, line 41: @dearrudam Is it possible to make `MappedWithIndex` a compositon of `Mapped` and a special `BiFunc`?

andreoss (changes requested):

src/main/java/org/cactoos/scalar/ScalarWithIndex.java
dearrudam: You're right @andreoss! I didn't realize that there's the `map.MapEntry` class that could replace such class without problems.

src/main/java/org/cactoos/iterator/MappedWithIndex.java
dearrudam: I didn't it because I've tried to be less intrusive to the existent components so I isolated the required pieces, but I think that I could change the `Mapped`'s constructor that accepts a `Func` to wraps it into a `BiFunc` then passing it to a new  `Mapped`'s constructor that accepts a BiFunc. What do you think about it?

dearrudam: @andreoss I've changed to follow your suggestion! 
You're right! These changes have provided a reusing of the components and decreasing of the number of classes to be maintained! Great!

src/main/java/org/cactoos/scalar/ScalarWithIndex.java
dearrudam: With my latest changes, we no longer need this class and derived ones.

dearrudam (commented):
@andreoss could you review the changes that I did?

dearrudam (commented):
@andreoss sorry man, I forgot to mark the files before submitting the review request... could you review them?

src/main/java/org/cactoos/iterable/MappedWithIndex.java
```diff
@@ -0,0 +1,79 @@
...
+     * @param <X> Type of source item
+     */
+    public <X> MappedWithIndex(
+        final BiFunc<Integer, ? super X, ? extends Y> fnc,
+        final Iterable<? extends X> src
+    ) {
+        super(
+            new IterableOf<>(
+                () -> new org.cactoos.iterator.Mapped<>(
+                    new Func<X, Y>() {
```
andreoss, line 66: @dearrudam This part is duplicated. 
It should be possible to re-use `itererator.MappedWithIndex` from `iterable.MappedWithIndex`.

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * Test case for {@link MappedWithIndex}.
+ * @since 0.50
+ * @checkstyle JavadocMethodCheck (500 lines)
+ * @checkstyle MagicNumberCheck (500 lines)
+ */
+final class MappedWithIndexTest {
+    @Test
+    void transformsList() throws Exception {
+        MatcherAssert.assertThat(
```
andreoss, line 46: @dearrudam Please use `cactoos-matchers` instead of `hamcrest`  (i.e `Assertion` instead of `MatcherAssertc, `IsEqual` instead of `Matchers.is`, etc)

andreoss (changes requested):

dearrudam (commented):
@andreoss please, could you review again? thanks

src/main/java/org/cactoos/iterable/MappedWithIndex.java
```diff
@@ -0,0 +1,77 @@
...
+     * @param fnc Func
+     * @param src Source iterable
+     * @param <X> Type of source item
+     */
+    public <X> MappedWithIndex(
+        final BiFunc<Integer, ? super X, ? extends Y> fnc,
+        final Iterable<? extends X> src
+    ) {
+        super(
+            new Mapped<>(
```
andreoss, line 64: @dearrudam Why don't you use `iterator.MappedWithIndex` here?

andreoss (changes requested):

src/main/java/org/cactoos/iterable/MappedWithIndex.java
dearrudam: You’re right @andreoss ! My bad... I’ll fix it ASAP. Thanks man!

dearrudam (commented):
@andreoss it looks like that the Github Actions didn't work properly... it looks like some netwokk issues happened....anyway... please, could you review it again?

andreoss (approved):
@dearrudam Looks fine.
Feel free to report flaky tests you encountered.

dearrudam:
The test `io.InputOfTest.readsSecureUrlContent` consumes the `https://www.yegor256.com/robots.txt` URL... at the moment of the test, the `Tests (ubuntu-18.04, 11)` container couldn't access there than broken the subsequent checks... IMHO I think that maybe it's needed to put some retry logic or maybe point to other robots.txt URLs (locally or fake or real) in order to avoid this kind of issue in the future.

src/main/java/org/cactoos/iterable/MappedWithIndex.java
```diff
@@ -0,0 +1,67 @@
...
+public final class MappedWithIndex<Y> extends IterableEnvelope<Y> {
+    /**
+     * Ctor.
+     * @param fnc Func
+     * @param src Source iterable
+     * @param <X> Type of source item
+     */
+    @SafeVarargs
+    public <X> MappedWithIndex(
+        final BiFunc<Integer, ? super X, ? extends Y> fnc,
```
victornoel, line 45: @dearrudam in terms of API niceness, I think it would be best to have a `BiFunc<? super X, Integer, ? extends Y>`

src/main/java/org/cactoos/iterable/MappedWithIndex.java
```diff
@@ -0,0 +1,67 @@
...
+
+import org.cactoos.BiFunc;
+
+/**
+ * Mapped with index iterable.
+ *
+ * <p>
+ * There is no thread-safety guarantee.
+ * @param <Y> Type of target item
+ * @since 0.50
```
victornoel, line 34: @dearrudam let's use `@since 1.0.0` on any new class

src/main/java/org/cactoos/iterator/MappedWithIndex.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Ctor.
+     * @param func Func
+     * @param iterator Source iterator
+     * @param <X> Type of item
+     * @checkstyle AnonInnerLengthCheck (60 lines)
```
victornoel, line 45: @dearrudam to avoid this checkstyle exception and the unnice anonymous class, what we can do is the following:
- introduce a primary private constructor that takes the `AtomicInteger` and the bifunc+iterator 
- then we can use `Mapped` with a simple lambda for the func in it
- the current consturctor becomes a public secondary one that gives the `new AtomicInteger(-1)`

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,109 @@
...
+import org.hamcrest.Matchers;
+import org.hamcrest.core.IsEqual;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.llorllale.cactoos.matchers.Assertion;
+
+/**
+ * Test case for {@link MappedWithIndex}.
+ * @since 0.50
+ * @checkstyle JavadocMethodCheck (500 lines)
```
victornoel, line 39: @dearrudam this exception is not needed

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,109 @@
...
+ * @checkstyle MagicNumberCheck (500 lines)
+ */
+final class MappedWithIndexTest {
+    @Test
+    void transformsList() throws Exception {
+        new Assertion<>(
+            "Can't transform an iterable",
```
victornoel, line 46: @dearrudam we use "Must" in cactoos instead of "Can't" in assertion messages

victornoel (changes requested):
@dearrudam here are a few more comments

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel thanks! I'll fix it.

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel okay! thanks! I'll fix it.

src/main/java/org/cactoos/iterator/MappedWithIndex.java
dearrudam: @victornoel thanks! I'll fix it! I'm learning a lot!

src/main/java/org/cactoos/iterable/MappedWithIndex.java
dearrudam: @victornoel okay! I'll fix it!

src/main/java/org/cactoos/iterable/MappedWithIndex.java
dearrudam: @victornoel thanks! I'm learning a lot!

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    void string() {
+        new Assertion<>(
+            "must convert to string",
+            new MappedWithIndex<>(
+                (x, index) -> x * index * 2,
+                new IterableOf<>(1, 2, 3)
+            ).toString(),
+            Matchers.equalTo("0, 4, 12")
```
victornoel, line 85: @dearrudam please avoid using any static method, here you can use `IsEqual`, and in `transformsEmptyList` you can directly throw an exception and also use new IterableOf<>() instead of `Collections.emptyList()`,  etc. You can also replace matchers such as `Matchers.emptyIterable()` with their equivalent objects instantiation with `new` (o use `IsEqual`). If some are unnice to use, don't hesitate to open new issues to cactoos-matchers to improve OO matchers :)

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,108 @@
...
+                        "-",
+                        new TextOf(index.toString()),
+                        new TextOf(input)
+                    )
+                ),
+                new IterableOf<>(
+                    "hello", "world", "друг"
+                )
+            ).iterator().next().asString(),
+            new IsEqual<>("0-HELLO")
```
victornoel, line 58: @dearrudam here I believe it's best to simply test equality to the fully built `Iterable`, not just one element, also let's try to keep tests as simple as needed to demonstrate that the object work: here I believe we don't need to combine upper + joined, some simpler string concatenate (with `Joined` or `Concatenated`) should be enough.

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,108 @@
...
+import org.llorllale.cactoos.matchers.Assertion;
+
+/**
+ * Tests for {@link MappedWithIndex}.
+ * @since 1.0.0
+ * @checkstyle MagicNumberCheck (500 lines)
+ */
+final class MappedWithIndexTest {
+    @Test
+    void transformsList() throws Exception {
```
victornoel, line 43: @dearrudam this actually should be called transformsIterable

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Test
+    void transformsArray() {
+        new Assertion<>(
+            "Transforms an array",
+            new MappedWithIndex<>(
+                (input, index) -> new Upper(
+                    new TextOf(
+                        String.format(
+                            "%1$s-%2$s",
```
victornoel, line 97: @dearrudam as above, let's use a simpler test to demonstrate the object behaviour. And if you need to format a string, it's best to use `FormattedText` btw, but a simple concatenation is good enough also

victornoel (changes requested):
@dearrudam some comments on the tests, I made them only on one of the file but they apply to both of course :)

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel I got it! You're right, I need to pay attention to these details! Thank you!

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel I'll fix it! Thanks!

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel I'll fix it! Thanks!

src/test/java/org/cactoos/iterable/MappedWithIndexTest.java
dearrudam: @victornoel I'll fix it! Thanks!

victornoel:
@dearrudam thx

victornoel:
@rultor merge

0crat:

sereshqua:
@dearrudam please make sure you start all of the comment with the name of the user they are refereed to, [see](https://github.com/yegor256/cactoos/pull/1554#discussion_r587519814)

sereshqua:
@0crat quality acceptable

Issue #1449, opened 2020-09-10
Some classes that perform Input/Output (LengthOf, BytesOf, ...) close the stream


fabriciofx:
Some classes as `LengthOf` closes the `Input` after performed the operation. But often we need keep the `Input` stream opened even after it. So, let's create a decorator to `Input` and `Output` that allows keep these streams opened.

victornoel:
@fabriciofx or we could also just not close those `Input` or `Output` in any class and introduce some concept to wrap operations that manipulates `Input`/`Output` to close them at the end (a bit like try-with-resource).

I'm not sure what is best…

fabriciofx:
@victornoel I think not close after the manipulation better. As say the golden rule: *"Who allocated the resource is the responsible to deallocated it". A try-with-resourced kind of class is a good idea too. :)

victornoel:
@fabriciofx actually, after some thought, I'm not sure, because when you call `Input.stream()`, it means you could retrieve a new stream, so you HAVE to close it at the end.

Then, it means your first proposal was actually best indeed: i.e., it is the `Input` object creator that decides how the streams returned from `stream` will behave.

victornoel:
@0crat in

victornoel:
@fabriciofx some extra requirements for this job for the DEV that will take it: we must find all the classes that closes the stream like this, document them about this behaviour and refer to the existence of the new classes we should introduce that prevent stream closing

baudoliver7:
@victornoel Let's me find all the classes that close the stream like mentioned by @fabriciofx, document them properly and add a todo. Isn't it ?

victornoel:
@baudoliver7 you can do it either order you want yes: first document then todo for implementing the new decorators, or first implement the new decorators and then todo to document the classes.

baudoliver7:
@victornoel Ok. Let's us implement the new decorators :)

baudoliver7:
@victornoel What names could you suggest me ?

victornoel:
@baudoliver7 hehe, no idea actually, @fabriciofx would you have an idea?

Could you propose a name and we discuss it in the PR maybe?

baudoliver7:
@victornoel Yes, I will do it.

andreoss:
@fabriciofx 
Keeping `OutputStream` opened might be handy, but what is the point of doing so for `InputStream`?

[...]

PR #1556, opened 2021-03-06
(#1449) Introduce new decorators to prevent stream to be closed

baudoliver7:
For #1449

- Introduce for `Input` decorators `SafeInput` and `SafeInputStream`
- Introduce for `Output` decorators `SafeOutput` and `SafeOutputStream`
- Introduce some unit tests for these new classes


andreoss:
@baudoliver7 
How would one actually close such stream? Unclosable streams will cause leakage. 
I don't think this solution is good. 

Instead I would
1) make sure that all calls to `Input#stream`/``Output#stream` is inside `try-with-resource`
2) provide additional methods (or smart-class) to avoid direct usage `try-with-resource`
`
```
interface Input {
     ....
    <X> X read(final Func<InputStream, ? extends X> func);
}
interface Output {
     ....
    void write(final Proc<OutputStream> proc);
}
```
For example
```
final Text text = new InputOf(...).read(in -> new TextOf(in));  // close is called by Input
```

victornoel:
@andreoss I think what you are proposing is out of scope of the original ticket, if you think the solution proposed there, maybe we can discuss it with @fabriciofx that proposed the original idea.

But  basically, the tests introduced by @baudoliver7 should answer your original question: sometimes you hold an input but you don't want that reading it with one of the existing class close it directly: you want to control as to when the closing happen.

A typical example of this could be reading a zip stream where you read each zipped file from the stream (Zip stream in the JRE are one stream for multiple files).

baudoliver7:
@andreoss WDYT about what said @victornoel ?

src/main/java/org/cactoos/io/SafeInput.java
```diff
@@ -0,0 +1,57 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+package org.cactoos.io;
+
+import java.io.InputStream;
+import org.cactoos.Input;
+
+/**
+ * A decorator of {@link Input} that provides a copy of {@link InputStream}
```
andreoss, line 30: @baudoliver7 "a copy of InputStream" can be misleading here.

src/test/java/org/cactoos/io/FakeInputStream.java
```diff
@@ -0,0 +1,65 @@
...
+ * Fake {@link InputStream} with ability to check if
+ * it is closed.
+ * @since 1.0.0
+ */
+final class FakeInputStream extends InputStream {
+
+    /**
+     * If {@link InputStream} is closed.
+     */
+    private boolean closed;
```
andreoss, line 39: @baudoliver7 It would be better as `AtomicBoolean`

src/test/java/org/cactoos/io/SafeOutputStreamTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.llorllale.cactoos.matchers.Assertion;
+
+/**
+ * Test case for {@link SafeOutputStream}.
+ * @since 1.0.0
+ */
+final class SafeOutputStreamTest {
+
+    @Test
+    @SuppressWarnings("try")
```
andreoss, line 38: @baudoliver7 
I would add more tests for reading and writing actual data using decorated streams.

andreoss (changes requested):

baudoliver7:
@andross You are right. `Stream` decorators are now `non public`.

baudoliver7:
@andreoss We can also read an `SafeInput` stream twice now :)

baudoliver7:
@andreoss Please, see my new changes.

victornoel:
@andreoss @baudoliver7 I think this went a bit out of hand ^^ the original design was correct: those decorators are not here to add caching or anything like that, they are just present to give back control of when the streams are closed to the caller instead of the code that I consuming the stream.

baudoliver7:
@victornoel Got it ! Let's me go back to previous design.

baudoliver7:
@victornoel @andreoss 
Previous design has been restored after having better understanding from `ARC` explanation. However, I keep the other suggestions made by @andreoss.

@andreoss, please, see my new changes.

src/main/java/org/cactoos/io/SafeInput.java
```diff
@@ -0,0 +1,65 @@
...
+     * Origin.
+     */
+    private final Input origin;
+
+    /**
+     * Ctor.
+     * @param stream Stream to preserve
+     */
+    public SafeInput(final InputStream stream) {
+        this(() -> stream);
```
victornoel, line 49: @baudoliver7 why introduce this constructor? I would expect that one can use `InputOf` if needed. Actually, building an `Input` based directly on an `InputStream` is not the most general type of implementation for inputs, because that makes them one-use.

victornoel (changes requested):

src/main/java/org/cactoos/io/SafeInput.java
baudoliver7: @victornoel I totally agree with you ! Let's remove this ctor. :)

baudoliver7:
@victornoel Please, see my changes.

victornoel:
@baudoliver7 looks good, but I would first like to get @andreoss point of view on all of this before merging

andreoss (approved):

victornoel:
@andreoss @baudoliver7 @fabriciofx now we need to decide if we are happy with `Safe` as a prefix :)
It's not bad but I wonder if we can't find anything better here..

victornoel:
@baudoliver7 @andreoss it's not easy... `CloseShieldInput` is interesting, `UncloseableInput` could work, or `NoncloseableInput`? `NoCloseInput`?

Not related but the way this `CloseShieldInputStream` class is implemented is interesting: the object behave as-is it was closed even if the underlying stream wasn't. This could be something we want to have... but well, let's wait for the need to arise :)

baudoliver7:
@victornoel Yes, `CloseShieldInput` is good one.

baudoliver7:
@victornoel Could I add this behaviour right now ?

victornoel:
@baudoliver7 if you like it, then let's go with that yes! thanks :)

baudoliver7:
@victornoel Could I apply `CloseShield` as suffix ?

baudoliver7:
@victornoel @andreoss Please review latest changes ?

src/main/java/org/cactoos/io/SafeInputStream.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    @Override
+    public int read(final byte[] buffer, final int offset,
+        final int length) throws IOException {
+        return this.inner.get().read(buffer, offset, length);
+    }
+
+    @Override
+    @SuppressWarnings("PMD.UncommentedEmptyMethodBody")
```
victornoel, line 71: @baudoliver7 we don't need this anymore then. Also, lets' remove the exception on the method since it's not actually thrown. Same for the output stream.

victornoel (changes requested):
@baudoliver7 one comment and yes, let's rename them as decided :)

src/main/java/org/cactoos/io/SafeInputStream.java
baudoliver7: @victornoel Right !

baudoliver7:
@victornoel Please, see new changes.

src/test/java/org/cactoos/io/CloseShieldInputStreamTest.java
```diff
@@ -0,0 +1,80 @@
...
+ * Test case for {@link CloseShieldInputStream}.
+ * @since 1.0.0
+ * @checkstyle ClassDataAbstractionCouplingCheck (500 lines)
+ */
+final class CloseShieldInputStreamTest {
+
+    @Test
+    @SuppressWarnings("try")
+    void preventsOriginalStreamToBeClosed() throws Exception {
+        try (FakeInputStream origin = new FakeInputStream()) {
```
andreoss, line 45: @baudoliver7 it should be `InputStream` on the left side

src/test/java/org/cactoos/io/CloseShieldOutputStreamTest.java
```diff
@@ -0,0 +1,81 @@
...
+                    new TextOf(temp)
+                )
+            ).affirm();
+        }
+    }
+
+    @Test
+    @SuppressWarnings("try")
+    void preventsOriginalStreamToBeClosed() throws Exception {
+        try (FakeOutputStream origin = new FakeOutputStream()) {
```
andreoss, line 70: @baudoliver7 Here it should be `OutputStream` on the left side, and in the class below as well

andreoss (changes requested):

src/test/java/org/cactoos/io/CloseShieldOutputStreamTest.java
baudoliver7: @andreoss Ditto.

src/test/java/org/cactoos/io/CloseShieldInputStreamTest.java
baudoliver7: @andreoss I'm afraid not because we need access to method #isClosed as below :
```java
try (FakeInputStream origin = new FakeInputStream()) {
    // @checkstyle EmptyBlockCheck (2 lines)
    try (InputStream stream = new CloseShieldInputStream(origin)) {
    }
    new Assertion<>(
        "Must not close origin stream",
        origin.isClosed(),
        new IsEqual<>(false)
    ).affirm();
}
```

baudoliver7:
@andreoss Please, see my answers.

src/test/java/org/cactoos/io/CloseShieldInputStreamTest.java
andreoss: @baudoliver7 I see now, sorry

andreoss (approved):

victornoel:
@rultor merge

0crat:

0crat:
Code review was too long (15 days), architects (@victornoel) were penalized, see [§55](http://www.zerocracy.com/policy.html#55)

sereshqua:
@0crat quality good

PR #1603, opened 2021-05-04
Make `Random` a parameter for `Shuffled`

andreoss:
- Shuffle with a specific randomizer

### DIFF
--- a/src/main/java/org/cactoos/iterable/Shuffled.java
+++ b/src/main/java/org/cactoos/iterable/Shuffled.java
@@ -48,6 +48,14 @@ public final class Shuffled<T> extends IterableEnvelope<T> {
 
     /**
      * Ctor.
+     * @param src The underlying iterable
+     */
+    public Shuffled(final Iterable<T> src) {
+        this(new SecureRandom(), src);
+    }
+
+    /**
+     * Ctor.
      * @param rnd Randomizer.
      * @param src The underlying iterable
      */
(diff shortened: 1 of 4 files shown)


src/main/java/org/cactoos/iterable/Shuffled.java
```diff
@@ -45,14 +48,26 @@ public Shuffled(final T... src) {
...
             )
         );
     }
 
+    /**
+     * Ctor.
+     * @param src The underlying iterable
+     */
+    public Shuffled(final Iterable<T> src) {
+        this(new SecureRandom(), src);
```
victornoel, line 70: @andreoss let's have the primary constructor last in the class, so let's move this constructor before the other one

src/main/java/org/cactoos/iterator/Shuffled.java
```diff
@@ -47,23 +49,29 @@
...
             )
         );
     }
 
+    /**
+     * Ctor.
+     * @param iterator The original iterator
+     */
+    public Shuffled(final Iterator<T> iterator) {
+        this(new SecureRandom(), iterator);
```
victornoel, line 72: @andreoss same here, let's have the primary constructor last

victornoel (changes requested):
@andreoss thx, 2 small feedbacks in the comments

andreoss:
@victornoel Done

PR #1608, opened 2021-05-10
(#1567) Generify `org.cactoos.proc` package

DmitryBarskov:
For #1567:

Generify incoming parameters (use `? super T`) for next classes:
- `ProcOf`, `ProcEnvelope` and `FuncOf`
- `ProcNoNulls`
- `BiProcOf` and `BiProcEnvelope`
- `BiProcNoNulls`
- `RepeatedProc` and `Repeated`

PR #1609, opened 2021-05-11
Create `BiProc` from two `Proc`

andreoss:
- Add ctor for creating `BiProc` from two `Procs`

### DIFF
--- a/src/main/java/org/cactoos/proc/BiProcOf.java
+++ b/src/main/java/org/cactoos/proc/BiProcOf.java
@@ -56,9 +56,22 @@ public final class BiProcOf<X, Y> extends BiProcEnvelope<X, Y> {
      * @param func The bi function
      */
     public BiProcOf(final BiFunc<? super X, ? super Y, ?> func) {
+        this((BiProc<X, Y>) func::apply);
+    }
+
+    /**
+     * Ctor.
+     * @param first The procedure on first arg.
+     * @param second The procedure on second arg.
+     */
+    public BiProcOf(
+        final Proc<? super X> first,
+        final Proc<? super Y> second
+    ) {
         this(
-            (x, y) -> {
-                func.apply(x, y);
+            (xvl, yvl) -> {
+                first.exec(xvl);
+                second.exec(yvl);
             }
         );
     }
@@ -69,8 +82,8 @@ public final class BiProcOf<X, Y> extends BiProcEnvelope<X, Y> {
      */
     public BiProcOf(final Proc<? super X> prc) {
         this(
-            (x, y) -> {
-                prc.exec(x);
+            prc,
+            yvl -> {
             }
         );
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/cactoos/proc/BiProcOf.java
```diff
@@ -56,9 +56,22 @@ public BiProcOf(final Func<? super X, ?> func) {
      * @param func The bi function
      */
     public BiProcOf(final BiFunc<? super X, ? super Y, ?> func) {
+        this((BiProc<X, Y>) func::apply);
```
victornoel, line 59: @andreoss if we can avoid, it would be best not to use casting, it would be best to use `(x,y) -> { func.apply(x,y); }` for example

src/main/java/org/cactoos/proc/BiProcOf.java
```diff
@@ -69,8 +82,8 @@ public BiProcOf(final BiFunc<? super X, ? super Y, ?> func) {
      */
     public BiProcOf(final Proc<? super X> prc) {
         this(
-            (x, y) -> {
-                prc.exec(x);
+            prc,
+            yvl -> {
```
victornoel, line 86: @andreoss please move this constructor above the one it calls

src/test/java/org/cactoos/proc/BiProcOfTest.java
```diff
@@ -130,4 +132,25 @@ void worksWithBiProc() throws Exception {
...
     }
+
+    @Test
+    void worksWithTwoProcs() throws Exception {
+        final AtomicInteger fst = new AtomicInteger();
+        final AtomicInteger snd = new AtomicInteger();
+        new Assertion<>(
+            "Must execute BiProc with two Procs",
+            new BiProcOf<>(
+                new BiProcOf<>(
```
victornoel, line 143: @andreoss why `BiProcOf` wrapped into a `BiProcOf`? Just one is enough, no?

victornoel (changes requested):

src/main/java/org/cactoos/proc/BiProcOf.java
andreoss: @victornoel Done

src/main/java/org/cactoos/proc/BiProcOf.java
andreoss: @victornoel Done

src/test/java/org/cactoos/proc/BiProcOfTest.java
andreoss: @victornoel Done

victornoel (approved):
@andreoss looks good, but unfortunately, when #1608 will be merged, there will be conflicts :)

victornoel:
@andreoss that was fast, thx :)

victornoel:
@rultor merge

victornoel:
@0crat status

Issue #1575, opened 2021-03-14
package-info.java:29-31: Continue to exploit generic...


0pdd:
The puzzle `1533-d1b82973` from #1533 has to be resolved: 

https://github.com/yegor256/cactoos/blob/297f62637cae056e9dd9e68886c45c985d5fc295/src/main/java/org/cactoos/collection/package-info.java#L29-L31

The puzzle was created by Olivier B. OURA on 08-Mar-21. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

victornoel:
@0crat status

0crat:
The job [#1575](https://github.com/yegor256/cactoos/issues/1575) assigned to @rocket-3[/z](https://www.0crat.com/u/rocket-3), here is [why](https://www.0crat.com/footprint/C63314D6Z/806f64f1-cb34-4fb1-a59f-aa9d6881ab66); the budget is 30 minutes, see [§4](http://www.zerocracy.com/policy.html#4); please, read [§8](http://www.zerocracy.com/policy.html#8) and [§9](http://www.zerocracy.com/policy.html#9); if the task is not clear, read [this](http://www.yegor256.com/2015/02/16/it-is-not-a-school.html) and [this](http://www.yegor256.com/2015/01/15/how-to-cut-corners.html); there will be no monetary reward for this job

0pdd:
The puzzle `1533-d1b82973` has disappeared from the source code, that's why I closed this issue.

victornoel:
@0crat status

victornoel:
@0crat status

victornoel:
@0crat status

victornoel:
@0crat status

victornoel:
@0crat status

PR #1624, opened 2021-09-07
(#1575) Generify `org.cactoos.collection` package

rocket-3:
For #1575.

Generify parameter types in constructors of org.cactoos.collection package classses:

- Immutable
- NoNulls

It's done by using unchecked casts, because there are absolutely no other ways to achive this, due to JDK interfaces limitations.

### DIFF
--- a/src/main/java/org/cactoos/collection/Immutable.java
+++ b/src/main/java/org/cactoos/collection/Immutable.java
@@ -25,7 +25,6 @@ package org.cactoos.collection;
 
 import java.util.Collection;
 import java.util.Iterator;
-import org.cactoos.iterable.IterableOf;
 
 /**
  * Decorator that doesn't allow any mutation of the wrapped {@link Collection}.
@@ -71,7 +70,7 @@ public final class Immutable<X> implements Collection<X> {
 
     @Override
     public Iterator<X> iterator() {
-        return new IterableOf<X>(this.col.iterator()).iterator();
+        return new org.cactoos.iterator.Immutable<>(this.col.iterator());
     }
 
     @Override
--- a/src/main/java/org/cactoos/collection/Immutable.java
+++ b/src/main/java/org/cactoos/collection/Immutable.java
@@ -54,8 +54,8 @@ public final class Immutable<X> implements Collection<X> {
      * Ctor.
      * @param src Source collection
      */
-    public Immutable(final Collection<X> src) {
-        this.col = src;
+    public Immutable(final Collection<? extends X> src) {
+        this.col = (Collection<X>) src;
     }
 
     @Override
--- a/src/main/java/org/cactoos/collection/Immutable.java
+++ b/src/main/java/org/cactoos/collection/Immutable.java
@@ -25,6 +25,7 @@ package org.cactoos.collection;
 
 import java.util.Collection;
 import java.util.Iterator;
+import org.cactoos.iterable.IterableOf;
 
 /**
  * Decorator that doesn't allow any mutation of the wrapped {@link Collection}.
@@ -48,15 +49,14 @@ public final class Immutable<X> implements Collection<X> {
     /**
      * Original collection.
      */
-    private final Collection<X> col;
+    private final Collection<? extends X> col;
 
     /**
      * Ctor.
      * @param src Source collection
      */
-    @SuppressWarnings("unchecked")
     public Immutable(final Collection<? extends X> src) {
-        this.col = (Collection<X>) src;
+        this.col = src;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


src/main/java/org/cactoos/collection/Immutable.java
```diff
@@ -54,8 +54,9 @@
      * Ctor.
      * @param src Source collection
      */
-    public Immutable(final Collection<X> src) {
-        this.col = src;
+    @SuppressWarnings("unchecked")
+    public Immutable(final Collection<? extends X> src) {
+        this.col = (Collection<X>) src;
```
victornoel, line 59: @rocket-3 instead of this, you should change the type of the `col` field. You won't need any cast in that case.

victornoel (changes requested):

src/main/java/org/cactoos/collection/Immutable.java
victornoel: @rocket-3 then you need to propagate the same kind of change everywhere needed until everything types correctly :) that's the purpose of this series of tickets: that the whole codebase uses generic with variance like this.

rocket-3: @victornoel, sorry, I didn't get it 

As I see, we have two ways:

- to do unchecked cast
- to rely on class which does uchecked cast

We can't propagate changes to JDK

victornoel: @rocket-3 We should be able to propagate as much as possible in our classes, and then normally, no cast is needed. If some casts are needed, let's put them at the last moment possible (usually inside a method implementation).

So let's go with this and see what happens with the two classes you started modifying!

victornoel: @rocket-3 by the way, it looks like `iterator()` is incorrectly implemented here since it should return an immutable iterator. In that case, I think no casting will be needed

rocket-3: @victornoel, no other way for `NoNulls`, cause we have an `add` and `addAll` impls.
We can't add items to a Collection<? extends X> by JDK design, only extract, see [PECS trilemma](https://stackoverflow.com/a/4343547/16625980). 

By the way, changed `Immutable`, see in the recent commit.

src/main/java/org/cactoos/collection/Immutable.java
```diff
@@ -70,7 +71,7 @@ public boolean isEmpty() {
 
     @Override
     public Iterator<X> iterator() {
-        return this.col.iterator();
+        return new IterableOf<X>(this.col.iterator()).iterator();
```
victornoel, line 74: @rocket-3 please use `iterator.Immutable` here since we must return an immutable iterator.

src/main/java/org/cactoos/collection/Immutable.java
victornoel: @rocket-3 ok, in that case, for `NoNulls`, because of its nature (being mutable), it's not possible to apply variance like this and thus the constructor should only take a X. If we cast, this will create errors at runtime if incorrect elements are added to such a collection. Let's also add a comment in the javadoc of the constructor so that we remember why it's like this.

rocket-3: @victornoel, wrote a test for this behaviour. Seems like no runtime exceptions, WDYT?

victornoel: @rocket-3 here is a failing test:
```java
    @Test
    void fails() {
        Collection<Double> a = new ArrayList<Double>();
        Collection<Number> b = new NoNulls<>(a);
        b.add(Long.valueOf(4));
        Double oupsNotADouble = a.iterator().next();
    }
```

This will throw `java.lang.ClassCastException: java.lang.Long cannot be cast to java.lang.Double` even though it compiles correctly.

rocket-3: Okay, I've removed uchecked cast, added impl. note and removed test created recently.

victornoel:
@rocket-3 nice use of `@implNote`, I didn't know about this!

victornoel (approved):

victornoel:
@rocket-3 the linters CI job failed, it must be fixed before we can merge this PR :)

rocket-3:
@victornoel, may be there's no need to fix that too? As in the https://github.com/yegor256/cactoos/pull/1627#issuecomment-922332593

rocket-3:
> @rocket-3 nice use of `@implNote`, I didn't know about this!

Also because of that `linters` CI job failed...

victornoel:
@rultor merge

victornoel:
@rocket-3 thx

PR #1626, opened 2021-09-12
Issue 1569 generify scalar package

rocket-3:
For #1569.

Relaxed parameter types by adding wildcards in constructors of classes of `org.cactoos.scalar` package:

- `And`
- `AndInThreads`
- `AndWithIndex`
- `FirstOf`
- `IoChecked`
- `ItemAt`
- `NoNulls`
- `PropertiesOf`
- `Repeated`
- `Retry`
- `ScalarOfSupplier`
- `Solid`
- `Sticky`
- `Synced`
- `Ternary`
- `Unckecked`
- `Xor`

### DIFF
--- a/src/main/java/org/cactoos/scalar/ItemAt.java
+++ b/src/main/java/org/cactoos/scalar/ItemAt.java
@@ -28,6 +28,7 @@ import java.util.Iterator;
 import org.cactoos.Func;
 import org.cactoos.Scalar;
 import org.cactoos.func.FuncOf;
+import org.cactoos.iterable.IterableOf;
 import org.cactoos.text.FormattedText;
 
 /**
@@ -53,7 +54,7 @@ public final class ItemAt<T> implements Scalar<T> {
      * @param position Position
      * @param iterable Iterable
      */
-    public ItemAt(final int position, final Iterable<T> iterable) {
+    public ItemAt(final int position, final Iterable<? extends T> iterable) {
         this(
             position,
             itr -> {
@@ -78,7 +79,7 @@ public final class ItemAt<T> implements Scalar<T> {
     public ItemAt(
         final int position,
         final T fallback,
-        final Iterable<T> iterable
+        final Iterable<? extends T> iterable
     ) {
         this(position, new FuncOf<>(new Constant<>(fallback)), iterable);
     }
@@ -92,8 +93,8 @@ public final class ItemAt<T> implements Scalar<T> {
      */
     public ItemAt(
         final int position,
-        final Func<Iterable<T>, T> fallback,
-        final Iterable<T> iterable
+        final Func<? super Iterable<? super T>, ? extends T> fallback,
+        final Iterable<? extends T> iterable
     ) {
         this.saved = new Sticky<T>(
             () -> {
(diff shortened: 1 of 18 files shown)


victornoel:
@0crat status

victornoel:
@0crat status

src/main/java/org/cactoos/scalar/ItemAt.java
```diff
@@ -92,8 +93,8 @@ public ItemAt(
      */
     public ItemAt(
         final int position,
-        final Func<Iterable<T>, T> fallback,
-        final Iterable<T> iterable
+        final Func<? super Iterable<? super T>, ? extends T> fallback,
```
victornoel, line 96: @rocket-3 I think this should be `Func<? super Iterable<? extends T>, ? extends T>`, i.e., with `extends` for the `Iterable` type parameter.

src/main/java/org/cactoos/scalar/package-info.java
```diff
@@ -26,8 +26,8 @@
  * Scalars.
  *
  * @since 0.12
- * @todo #1533:30min Exploit generic variance for package org.cactoos.scalar
- *  to ensure typing works as best as possible as it is explained in
- *  #1533 issue.
+ * @todo #1569:90min Create tests for the semantics of relaxed wildcards
```
victornoel, line 29: @rocket-3 just leave `30min` instead of 90, for two reasons: 1) we don't really use this info in practice and 2) in spirit, we always aim for micro tasks that take less than 30mn to encourage the use of todos and iterative development

victornoel (changes requested):
@rocket-3 just a few comments. Also, one last check: have you verified that the `scalar` package is completely covered with those relaxed generics?

rocket-3:
@0crat status

rocket-3:
> @rocket-3 just a few comments. Also, one last check: have you verified that the `scalar` package is completely covered with those relaxed generics?

@victornoel Yes, it's about all the package. If you're going to trust me, I've checked it again yesterday.

rocket-3:
@0crat status

victornoel:
@rocket-3 I do trust you, my question is just a "protocol" check just to ensure you checked it, not to doubt that you did ^^ that's why we have reviews :)

victornoel:
@rocket-3 so what about #1628 and the comment I made there: https://github.com/yegor256/cactoos/pull/1628#issuecomment-927296249 ? If the present PR is the one you want to keep, then please close the other

rocket-3:
@victornoel It (#1628) was just the same, I've closed it.

victornoel:
@rocket-3 even when running mvn manually with Java 8? I don't believe it's related to the settings.xml file (which is only there for credentials used by rultor). This is a typical problem with varargs and generics in Java, just add suppress warnings for those tests, as in here: https://github.com/yegor256/cactoos/blob/9f33693c15e48e10e7e5a62bbb32c66f26200c62/src/test/java/org/cactoos/set/SetOfTest.java#L46

I'm a bit surprised that the CI didn't complain though... but well, let's go with it and add that fix to the corresponding tests :)

rocket-3:
@victornoel Please, try to merge now

victornoel:
@rultor merge

victornoel:
@rocket-3 👍

rocket-3:
@0crat, status

PR #1629, opened 2021-10-07
Bump Qulice version to 0.19.0

rocket-3:
The new version seems to be working fine, maybe use it in the project too?


victornoel:
@rocket-3 this is a good idea, thx!

PR #1646, opened 2022-07-18
Added a note in readme.md

rocket-3:
Its about cactoos versions needed to be kept up-to-date by maintainers of library projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ Java version required: 1.8+.
 
 StackOverflow tag is [cactoos](https://stackoverflow.com/questions/tagged/cactoos).
 
+**Note:** please, don't forget to update `cactoos` version in your libraries, cause newer versions may break backwards compatibility and projects with newer version of `cactoos` may fail to even compile with your library then (due to Maven uses the newest version only for the whole compile context).
+
 ## Input/Output
 
 More about it here:


victornoel:
@rocket-3 do we really need this? I mean, this a 0.x versioned project for now, it's kind of part of the deal.

Or instead, please rephrase the text to focus on the idea that this project is still in development, that breaking changes happens from time to time and that v1.0.0 will be stabilized. You can even link to the milestone for 1.0.0 :)

Thx!

README.md
```diff
@@ -59,6 +59,8 @@ Java version required: 1.8+.
 
 StackOverflow tag is [cactoos](https://stackoverflow.com/questions/tagged/cactoos).
```
edmoffo, line 60: Maven's transitive dependency mediation resolves to the nearest declaration in the dependency tree, not the newest. See https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#transitive-dependencies. The parenthetical here states the mechanism backwards and undercuts the rest of the note; drop the parenthetical or replace it with the accurate rule.

edmoffo (changes requested):
One inline comment on the parenthetical about Maven's resolution mechanism, which states the rule backwards and undercuts the advice.

yegor256:
Merged, thanks!

---

PR #1676, opened 2023-01-22
Remove unnecessary ignores

alex-semenyuk:
Remove unnecessary ignores. Currently it seems as expected result

### DIFF
--- a/src/test/java/org/cactoos/map/NoNullsTest.java
+++ b/src/test/java/org/cactoos/map/NoNullsTest.java
@@ -30,7 +30,6 @@ import org.hamcrest.MatcherAssert;
 import org.hamcrest.core.AllOf;
 import org.hamcrest.core.IsEqual;
 import org.hamcrest.core.IsIterableContaining;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.llorllale.cactoos.matchers.Assertion;
 import org.llorllale.cactoos.matchers.IsEntry;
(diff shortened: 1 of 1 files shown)


src/test/java/org/cactoos/map/NoNullsTest.java
```diff
@@ -297,8 +295,7 @@ public void removeWithNullKey() {
         );
     }
 
-    @Test
-    @Ignore
+    @Test(expected = IllegalStateException.class)
```
alex-semenyuk, line 298:
```
        final V result = this.map.remove(key);
        if (result == null) {
            throw new IllegalStateException(
                new UncheckedText(
                    new FormattedText(
                        "Value returned by #remove(%s) is NULL",
                        key
                    )
                ).asString()
            );
        }
```

src/test/java/org/cactoos/map/NoNullsTest.java
```diff
@@ -251,8 +250,7 @@ public void putWithNullValue() {
...
```
alex-semenyuk, line 253: final V result = this.map.put(key, value);
        if (result == null) {
            throw new IllegalStateException(
                new UncheckedText(
                    new FormattedText(
                        "Value returned by #put(%s,%s) is NULL",
                        key, value
                    )
                ).asString()
            );
        }

victornoel:
Hi @alex-semenyuk, I'm not exactly sure what are the meaning of the comments you added… can you enlighten me?

Thank you for the PR :)

alex-semenyuk:
Just to show from where the result come from and that it's expected

PR #1689, opened 2023-08-20
reentrant lock

### LINKED ISSUE #1688: Bad performance locking
actually this code is not parallel at all because of the `synchronized` keyword, I suggest using `reentrant lock`, wdyt?

https://github.com/yegor256/cactoos/blob/9ee0dd1fef855129104fdbbf6e66d6eafd4bfb05/src/main/java/org/cactoos/scalar/Synced.java#L82

l3r8yJ:
closes #1688 

@fabriciofx take a look, please

### DIFF
--- a/src/main/java/org/cactoos/scalar/Synced.java
+++ b/src/main/java/org/cactoos/scalar/Synced.java
@@ -24,6 +24,7 @@
 package org.cactoos.scalar;
 
 import org.cactoos.Scalar;
+import java.util.concurrent.locks.ReentrantLock;
 
 /**
  * Scalar that is thread-safe.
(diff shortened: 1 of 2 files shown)


fabriciofx:
@l3r8yJ 

please, change the constructors to something like this:

```
public Synced(final Scalar<? extends T> scalar) {
   this(scalar, new ReetrantLock());
}

public Synced(final Scalar<? extends T> scalar, final ReentrantLock lock) {
   this.origin = scalar;
   this.lock = lock;
}
```

(and the same to `Synced` in `Text`.

Another thing: can you provide a Test to measure the performance improvement (because now is parallel)?

victornoel:
Hello @l3r8yJ, what is the advantage of using this over the synchronized keyword? synchronized and reentrant locks are exactly the same semantically as far as I know.

The only difference is when using the new Java 21 virtual threads which does not support (yet) the synchronized keyword.

yegor256:
Not merging this. Two concerns: CI never actually ran on the head commit (status is still "pending" with zero reported checks after three years), so there's no green build to point to; and the diff removes the public `Synced(Scalar, Object)` and `Synced(Text, Object)` constructors entirely rather than adding the `ReentrantLock` alongside them, which is a breaking change for anyone passing their own lock object. Given the branch's age it's also likely to need a rebase against current master.

Reassigning to yegor256 for manual review.

---

yegor256:
Please forgive us for closing this pull request — it has had merge conflicts with `master` for over a month, and per our policy we close stale conflicting PRs. Feel free to submit a new one, perhaps rebased against the current `master`, taking into account whatever changed there since this was opened. Thank you for the contribution and sorry for the inconvenience.

PR #1692, opened 2023-08-30
ThrowsOnFalse

### LINKED ISSUE #1175: TrueOrThrow class
I have a use case that I think might be common enough to warrant creating a dedicated class in this library to handling. I need to evaluate a `Boolean` expression and throw an `Exception` if that expression is false. It should probably implement `Scalar<Boolean>` and have a signature of `TrueOrThrow(expression, exceptionMessage)`.

```java
new TrueOrThrow(
    () -> 1 == 0,    
    "1 does not equal 0."
).value();
```

For completeness, its counterpart would be:

```java
new FalseOrThrow(
    () -> 1 == 1,
    "1 does equal 1, but we need it not to for some reason."
).value();
```

h1alexbel:
closes #1175

### DIFF
--- a/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
+++ b/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
@@ -46,19 +46,6 @@ public final class ThrowsOnFalse implements Scalar<Boolean> {
     /**
      * Ctor.
      * @param sclr Scalar
-     * @param exc Exception
-     */
-    public ThrowsOnFalse(
-        final Scalar<Boolean> sclr,
-        final Scalar<Exception> exc
-    ) {
-        this.scalar = sclr;
-        this.exception = exc;
-    }
-
-    /**
-     * Ctor.
-     * @param sclr Scalar
      * @param message Error Message
      */
     public ThrowsOnFalse(
@@ -68,6 +55,19 @@ public final class ThrowsOnFalse implements Scalar<Boolean> {
         this(
             sclr, () -> new IllegalArgumentException(message)
         );
+    }
+
+    /**
+     * Ctor.
+     * @param sclr Scalar
+     * @param exc Exception
+     */
+    public ThrowsOnFalse(
+        final Scalar<Boolean> sclr,
+        final Scalar<Exception> exc
+    ) {
+        this.scalar = sclr;
+        this.exception = exc;
     }
 
     @Override
--- a/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
+++ b/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
@@ -39,26 +39,41 @@ public final class ThrowsOnFalse implements Scalar<Boolean> {
     private final Scalar<Boolean> scalar;
 
     /**
-     * Error message.
+     * Exception to throw.
      */
-    private final String message;
+    private final Scalar<Exception> exception;
 
     /**
      * Ctor.
      * @param sclr Scalar
-     * @param msg Message
+     * @param exc Exception
      */
-    public ThrowsOnFalse(final Scalar<Boolean> sclr, final String msg) {
+    public ThrowsOnFalse(
+        final Scalar<Boolean> sclr,
+        final Scalar<Exception> exc
+    ) {
         this.scalar = sclr;
-        this.message = msg;
+        this.exception = exc;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


h1alexbel:
@yegor256 take a look, please

yegor256:
@victornoel what do you think?

src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
```diff
@@ -0,0 +1,65 @@
...
+     * Error message.
+     */
+    private final String message;
+
+    /**
+     * Ctor.
+     * @param sclr Scalar
+     * @param msg Message
+     */
+    public ThrowsOnFalse(final Scalar<Boolean> sclr, final String msg) {
```
victornoel, line 51: @h1alexbel can you also add a constructor `ThrowsOnFalse(final Scalar<Boolean> sclr, final Scalar<Exception> exception)` to allow the user to choose which exception will be thrown

src/test/java/org/cactoos/scalar/ThrowsOnFalseTest.java
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * @since 0.56.0
+ */
+final class ThrowsOnFalseTest {
+
+    @Test
+    void throwsOnFalse() {
+        final String msg = "test message";
+        final String message =
+            Assertions.assertThrows(
```
victornoel, line 43: @h1alexbel please use the `Assertion` class to write assertions, see other tests for some examples.

victornoel (changes requested):
@h1alexbel thank you for your contribution, here are a few comments.

h1alexbel:
@victornoel fixed, check again, please

src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
```diff
@@ -39,26 +39,41 @@ public final class ThrowsOnFalse implements Scalar<Boolean> {
...
+     * @param sclr Scalar
+     * @param message Error Message
+     */
+    public ThrowsOnFalse(
+        final Scalar<Boolean> sclr,
+        final String message
+    ) {
+        this(
+            sclr, () -> new IllegalArgumentException(message)
+        );
```
victornoel, line 70: @h1alexbel please put the constructor in this order in the file: first secondary constructor (the one taking a String) and last primary constructor (the one taking the two scalars).

src/test/java/org/cactoos/scalar/ThrowsOnFalseTest.java
```diff
@@ -24,50 +24,38 @@
...
-                ).value()
-            ).getMessage();
-        MatcherAssert.assertThat(
-            "Exception message in right format",
-            message,
-            Matchers.equalTo(msg)
-        );
+    @Test(expected = IllegalArgumentException.class)
+    public void throwsOnFalse() throws Exception {
+        new ThrowsOnFalse(
```
victornoel, line 40: @h1alexbel please use `org.llorllale.cactoos.matchers.Throws` with `Assertion` to check for exception (instead of using `expected = IllegalArgumentException.class`)

src/test/java/org/cactoos/scalar/ThrowsOnFalseTest.java
```diff
@@ -24,50 +24,38 @@
...
+            () -> false, "test message"
```
victornoel, line 41: @h1alexbel also check for the error message itself, not only that the exception is thrown

victornoel (changes requested):

h1alexbel:
@victornoel fixed, take a look again, please

victornoel (approved):

victornoel:
@h1alexbel looks good, thank you :)

victornoel:
@rultor merge

victornoel:
@yegor256 not sure why we have to codacy checks, but they make no sense with how we write tests, the error is:
> JUnit tests should include assert() or fail() 

How can we disable them?

yegor256:
@victornoel disabled. Once something new is pushed to this branch, CI checks will be re-run (without codacy)

victornoel:
@h1alexbel please could you push again to this branch, for example squash all commits as one, it would be perfect.

0crat:
Job `gh:yegor256/cactoos#1692` is not assigned, can't get performer

PR #1696, opened 2023-09-11
Throw on false

### LINKED ISSUE #1175: TrueOrThrow class

h1alexbel:
@victornoel take a look, please
it's a squashed of #1692 
closes #1175

### DIFF
--- a/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
+++ b/src/main/java/org/cactoos/scalar/ThrowsOnFalse.java
@@ -1,0 +1,80 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2017-2022 Yegor Bugayenko
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


victornoel:
@h1alexbel hi, there was no need to create a different PR, you could have just force pushed to your previous branch. Anyway, now that it is done, it's good as it is, thank you :)

victornoel:
@rultor merge

0crat:
Job was finished in 0 hours, bonus for fast delivery is possible (see [§36](http://www.zerocracy.com/policy.html#36))

Issue #1704, opened 2023-12-09
Let's implement a Multiline Text


fabriciofx:
Let's implement a Multiline Text which gets a big text and breaks it in lines (according a limit of line length) and output a `Interable<Text>`.

@victornoel WDYT?

victornoel:
@fabriciofx so it would be ` new MultilineText(Text txt, int limit)` and the class implements `Iterable<Text>`, correct?

fabriciofx:
@victornoel Yes, correct. WDYT?

victornoel:
@fabriciofx sure, why not, I'm not sure if 0crat is still alive though, let's see…

victornoel:
@0crat in

PR #1706, opened 2023-12-28
(#1704) Implement a Multiline Text

fabriciofx:


src/test/java/org/cactoos/text/MultilineTest.java
```diff
@@ -0,0 +1,113 @@
...
+                    "The quick brown fox jumps over the lazy black dog",
+                    "and after that returned to the cave"
+                )
+            ),
+            new IsEqual<>(
+                new IterableOf<>(
+                    new Joined(
+                        " ",
+                        "The quick brown fox jumps over the lazy",
+                        "black dog and after that returned to "
```
pnatashap, line 106: If I can still calculate correctly, the string  'The quick brown fox jumps over the lazy black dog and after that returned to **the**'  is exactly 80 symbols which is the default value, so looks like expected value is wrong

fabriciofx:
@pnatashap Can you take another look, please?

pnatashap:
Now it works fine, several points that it is not easy to understand (just my point of view):
1. StringBuilder with line is current word or this word with space, not sure that we need separate builder for this.
2. Not very clear that we try to check current word and the next word in one iteration, it looks like can be simplified

pnatashap:
@yegor256 @victornoel please take a look

yegor256:
@fabriciofx what is the motivation for this? Do we have a ticket?

PR #1712, opened 2024-02-08
Add Strict for Scalar

kerelape:
Added `Strict` scalar, that encapsulates another scalar, a rule for validating that original scalar, and an exception, that is thrown in case the original value doesn't meet the rule.


kerelape:
@yegor256 @victornoel please, run the workflows

yegor256:
@kerelape some violations of Qulice

kerelape:
@yegor256 updated copyright year - must be OK now

yegor256:
@rultor merge

Issue #1722, opened 2024-03-30
missing analog for StringUtils.splitPreserveAllTokens


pnatashap:
Link to the documentation
https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#splitPreserveAllTokens-java.lang.String-char-

The difference to Split is that empty line is added in case if duplicated delimiter is presented.

yegor256:
Implemented: `org.cactoos.text.SplitPreserveAllTokens` is in `src/main/java/org/cactoos/text/SplitPreserveAllTokens.java` and keeps the empty tokens produced by adjacent delimiters. Closing as done.

PR #1743, opened 2024-11-01
 Adding an OOP analogue for StringUtils.splitPreserveAllTokens ISSUE #1722

tjann7:
Rationale of some decisions made:
1) Functional Interface allows use of custom String splitter function avoiding the static method calls
2) A new class created as modifying already existing Split class would not be shorter : if we add some checking for flags would create just as many constructors as with 2 distinct classes
3) Tests only compare lengths of obtained Collections

### DIFF
--- a/src/main/java/org/cactoos/func/TriFuncSplitPreserve.java
+++ b/src/main/java/org/cactoos/func/TriFuncSplitPreserve.java
@@ -29,7 +29,17 @@ import org.cactoos.TriFunc;
 import org.cactoos.list.ListOf;
 
 /**
- * Used for avoiding static method calls.
+ * A String splitter preserving all tokens.
+ * Unlike regular Split, stores empty "" tokens
+ * created by adjacent regex separators.
+ *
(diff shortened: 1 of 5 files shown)


src/main/java/org/cactoos/func/TriFuncSplitPreserve.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+
+package org.cactoos.func;
+
+import java.util.Collection;
+import org.cactoos.TriFunc;
+import org.cactoos.list.ListOf;
+
+/**
+ * Used for avoiding static method calls.
```
yegor256, line 32: @tjann7 maybe you can provide a more detailed explanation here, with an example? Currently, it's hard to understand what this class is for, exactly.

tjann7:
@yegor256 please review
Also, is it okay to tag people always asking for a review or you will reach eventually on your own?

yegor256:
@rultor merge

PR #1757, opened 2025-01-22
Added support for getting a list of files from a jar directory in ResourceOf.

PeJetuz:
Added support for getting file names when reading a directory from a jar.
Added a test for getting file names when reading a directory.
Added 'test-data' module that builds a test jar file.


PeJetuz:
@yegor256 Will you please review?

yegor256:
@PeJetuz thanks for the contribution! I'm not sure the location of the `test-data/` directory is correct. I believe, all test resources and data must stay inside `src/test/` directory, by convention in Maven.

PeJetuz:
@yegor256 Thanks! I don't like `test-data/` either. I thought maybe someone would like the original project to build `test-data-0.0.1.jar`. Since it only contains 3 text files, it should be easy to recreate.

yegor256:
@PeJetuz indeed, it's good to have sources in Git repo, instead of a binary `.jar`. But we should keep files in the `src/test/`, not in the root.

PeJetuz:
@yegor256 Now `test-data-0.0.1.jar` is in `src/test/resources/`. For testing, we need to have a separate jar file in the classpath. Then in tests we can get a list of files from this jar. Or I don't understand something, please explain.

yegor256:
@PeJetuz "force push" is a bad idea in general, because it's confusing for the reviewer (just FYI)

PeJetuz:
@yegor256 I thought it was important to you that the intermediate results were not included in the final commit. Thank you, I will fix it!

PeJetuz:
@yegor256 Done. Thanks!

src/main/java/org/cactoos/io/ResourceOf.java
```diff
@@ -240,4 +257,76 @@ public InputStream stream() throws Exception {
...
+            }
+            return
+                new InputStreamOf(
+                    names
+                        .stream()
+                        .collect(
+                            Collectors.joining("\n")
+                        )
+                        .getBytes(StandardCharsets.UTF_8)
+                );
```
yegor256, line 317: @PeJetuz why not this:

```
return new InputStreamOf(
  names
    .stream()
    .collect(Collectors.joining("\n"))
    .getBytes(StandardCharsets.UTF_8)
);
```

src/main/java/org/cactoos/io/ResourceOf.java
```diff
@@ -240,4 +257,76 @@ public InputStream stream() throws Exception {
...
+        public InputStream files() throws Exception {
+            final List<String> names = new ArrayList<>(2);
+            try (JarFile jar = new JarFile(this.extract())) {
+                final Enumeration<JarEntry> entries = jar.entries();
+                while (entries.hasMoreElements()) {
+                    final JarEntry entry = entries.nextElement();
+                    final String name = entry.getName();
+                    if (this.path.equals(name)) {
+                        continue;
+                    } else if (name.lastIndexOf(this.path) >= 0) {
```
yegor256, line 304: @PeJetuz read this, please: https://www.yegor256.com/2015/01/21/if-then-throw-else.html

yegor256:
@PeJetuz looks much better now, but I still suggest placing it into `src/test/` instead of `src/test/resources/`. All files from `src/test/resources/` are automatically moved to `target/test-classes/`. I believe, this is not your plan for these files.

src/main/java/org/cactoos/io/ResourceOf.java
PeJetuz: @yegor256 Done.

src/main/java/org/cactoos/io/ResourceOf.java
PeJetuz: @yegor256 Thanks for the article. I replaced while/if/else with stream.

PeJetuz:
@yegor256 I changed the location of test-lib to src/test.  I replaced while/if/else with stream and fixed the formatting.
Sorry, I didn't know force-push breaks changelog.
Thanks for your help!

PeJetuz:
@yegor256 Hi! Is there anything else that needs to be improved?

yegor256:
We are closing this pull request, because it has merge conflicts with `master` and has been open for more than a month. Please, forgive us for this: it says nothing about the quality of your work, it's just housekeeping.

If the change still matters to you, please submit a new pull request. It may be very similar to this one, but it has to take into account everything that landed in `master` since then.

Thank you for contributing, and sorry again for the inconvenience!

PR #1758, opened 2025-01-23
Improve NoNulls map code coverage

Ivruix:
Added tests that cover `put` and `remove` methods of `NoNulls` map class.


Ivruix:
@yegor256 please take a look

yegor256:
@Ivruix thanks!

PR #1767, opened 2025-02-21
Fix previousIndex method in JoinedListIterator

Ivruix:
Fixed an off by one error mentioned in #1762. Added test to ensure correct behavior.


Ivruix:
@yegor256 please take a look

Ivruix:
@yegor256 please take a look

yegor256:
@Ivruix thanks!

PR #1777, opened 2025-03-14
Fix incorrect end predicate in Sliced iterator's second constructor

alimansour0002:
## Description
This PR fixes an issue with the `Sliced` iterator class where the second constructor was using an incorrect end predicate.

### Problem
The current implementation uses `any -> !iterator.hasNext()` which captures the iterator's state at construction time. This creates a fixed boolean value that doesn't change when the actual `hasNext()` method is called later, leading to incorrect iteration behavior.

### Solution
Changed the predicate to `any -> false` which means "never end because of the predicate". The actual end condition is properly checked in the `hasNext()` method with `this.iterator.hasNext()`.

## Related Issue
Fixes #[issue-number]

### DIFF
--- a/src/main/java/org/cactoos/iterator/Sliced.java
+++ b/src/main/java/org/cactoos/iterator/Sliced.java
@@ -59,7 +59,7 @@ public final class Sliced<T> implements Iterator<T> {
      * @param iterator Decorated iterator
      */
     public Sliced(final int start, final Iterator<? extends T> iterator) {
-        this(start, any -> !iterator.hasNext(), iterator);
+        this(start, any -> false, iterator);
     }
 
     /**
--- a/src/test/java/org/cactoos/iterator/SlicedTest.java
+++ b/src/test/java/org/cactoos/iterator/SlicedTest.java
@@ -4,7 +4,11 @@
  */
 package org.cactoos.iterator;
 
+import java.util.Iterator;
+import java.util.LinkedList;
+import java.util.List;
 import java.util.NoSuchElementException;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.cactoos.iterable.IterableOf;
 import org.cactoos.scalar.Constant;
 import org.hamcrest.core.IsEqual;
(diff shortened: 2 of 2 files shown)


alimansour0002:
The issue with `any -> !iterator.hasNext()` is that it captures the iterator's state at construction time rather than when the predicate is evaluated. While this doesn't cause failures in typical usage (thanks to the `this.iterator.hasNext()` check in the `hasNext()` method), it's conceptually incorrect.

Changing to `any -> false` makes the code cleaner and more robust by:
Avoiding state capture at construction time
Making the predicate's intent clearer - "never end because of the predicate"
Ensuring behavior depends only on the current state of the underlying iterator

mapogolions:
Currently, if the user does not specify a count limit, the end predicate takes the form `any -> !iterator.hasNext()`. This results in duplicate calls, as reproduced by the attached unit test.

@alimansour0002
Feel free to use the provided unit test for your PR.

Therefore, when the user does not specify count, it seems more appropriate for the end predicate to take the form `any -> false`.


```java

 @Test void testYourName() {
        final List<String> list = new LinkedList<>();
        list.add("1");
        AtomicInteger counter = new AtomicInteger(0);
        Runnable hook = () -> counter.incrementAndGet();
        final Iterator<String> hooked = new Hooked<>(hook, list.iterator());
        final Iterator<String> sliced = new Sliced<>(0, hooked);
        sliced.hasNext();
        assertEquals(1, counter.get());
    }

    private final class Hooked<T> implements Iterator<T> {

        private final Runnable hook;
        private final Iterator<T> origin;

        public Hooked(final Runnable hook, final Iterator<T> origin) {
            this.hook = hook;
            this.origin = origin;
        }

        @Override
        public boolean hasNext() {
            hook.run();
            return this.origin.hasNext();
        }

        @Override
        public T next() {
            hook.run();
            return this.origin.next();
        }
    }

```

alimansour0002:
@yegor256 the added test `hasNextCallsOriginOnce`  fails for the previous implementation, while it passes now
@mapogolions

yegor256:
@rultor merge

yegor256:
@alimansour0002 thanks!

PR #1781, opened 2025-03-17
Implement Set Intersection functionality

alimansour0002:
This PR adds a new `Intersection` class to the cactoos library that provides set intersection functionality. The implementation follows the same pattern as the existing `Diff` class.

## Features
- Implements set intersection operation (elements common to both sets)
- Provides multiple constructors to handle different input types:
  - Iterable
  - Iterator
  - Scalar
  - SetOf
- Includes comprehensive test coverage

## Implementation Details
The `Intersection` class extends `SetEnvelope<T>` and uses Java's `retainAll` method to compute the intersection between two sets. This approach is consistent with the existing `Diff` class implementation.

## Test Coverage
The test class includes tests for:
- Basic intersection functionality
- Intersection with empty sets
- Intersection with Java's built-in Set implementation
- Intersection with iterables and iterators

This implementation completes the set of basic set operations (union is already provided by SetOf, difference by Diff, and now intersection) making the library more comprehensive for set manipulation tasks.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "java.configuration.updateBuildConfiguration": "interactive"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "java.configuration.updateBuildConfiguration": "interactive"
-}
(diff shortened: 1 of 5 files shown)


alimansour0002:
@yegor256 could you check this please

alimansour0002:
@yegor256

yegor256:
@alimansour0002 it's better to keep IDE-related files in `.gitignore`

yegor256:
@alimansour0002 thanks!

Issue #1852, opened 2026-02-12
Solid doesn't compute null value only once


fabriciofx:
@yegor256 `Solid` doesn't comput null value only once. For example, this code:

```java
final AtomicInteger calls = new AtomicInteger(0);
final Scalar<Object> scalar = new Solid<>(
    () -> {
        calls.incrementAndGet();
        return null;
    }
);
scalar.value();
scalar.value();
```

Will fail to test `calls.get() == 1` (it's 2).

yegor256:
@fabriciofx definitely a bug, thanks

PR #1855, opened 2026-02-22
#1852: Fix Solid does not compute null value only once

fabriciofx:
@yegor256 Fix Solid does not compute null value only once

### DIFF
--- a/src/test/java/org/cactoos/scalar/SolidTest.java
+++ b/src/test/java/org/cactoos/scalar/SolidTest.java
@@ -5,8 +5,10 @@
 package org.cactoos.scalar;
 
 import java.security.SecureRandom;
+import java.util.List;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.cactoos.Scalar;
+import org.cactoos.experimental.Threads;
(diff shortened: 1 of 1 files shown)


src/main/java/org/cactoos/scalar/Solid.java
```diff
@@ -44,10 +49,11 @@ public Solid(final Scalar<? extends T> origin) {
 
     @Override
     public T value() throws Exception {
-        if (this.cache == null) {
+        if (!this.computed) {
             synchronized (this.lock) {
-                if (this.cache == null) {
+                if (!this.computed) {
                     this.cache = this.origin.value();
+                    this.computed = true;
```
yegor256: @fabriciofx what about thread-safety, which is promised in this class?

yegor256:
@fabriciofx try to use 100 threads, you will see the problem

fabriciofx:
@yegor256 as you can see in the PR, I added a test to verify that my code works with 100 threads, and everything indicates that it does. If you have any other way to verify that it does not work with multiple threads, please let me know so I can fix my patch.

PR #1891, opened 2026-05-19
Replace StringBuffer with StringBuilder in org.cactoos.text.Replaced (#1410)

### LINKED ISSUE #1410: org.cactoos.text.Replaced uses StringBuffer instead of StringBuilder
The class `org.cactoos.text.Replaced` uses `StringBuffer` instead of `StringBuilder` ([here](https://github.com/yegor256/cactoos/blob/3675019c59c64477e8b80a04a25f86d717d9ac4b/src/main/java/org/cactoos/text/Replaced.java#L88)). Let's change it because `StringBuilder` is more efficient.

mvanhorn:
Closes #1410.

Replaces `StringBuffer` with `StringBuilder` in `org.cactoos.text.Replaced`. The buffer is scoped to a single lambda invocation and never escapes that thread, so the synchronization `StringBuffer` inherits from `AbstractStringBuilder` is wasted work.

`Matcher.appendReplacement(StringBuilder, String)` and `Matcher.appendTail(StringBuilder)` have been available since Java 9; the project's `pom.xml` already targets Java 17, so the StringBuilder overloads compile cleanly.

The existing `ReplacedTest` covers the public behaviour and stays unchanged.

### DIFF
--- a/src/main/java/org/cactoos/text/Replaced.java
+++ b/src/main/java/org/cactoos/text/Replaced.java
@@ -67,7 +67,7 @@ public final class Replaced extends TextEnvelope {
         super(
             new Mapped(
                 str -> {
-                    final StringBuffer buffer = new StringBuffer();
+                    final StringBuilder buffer = new StringBuilder();
                     final Matcher matcher = regex.value().matcher(str);
                     while (matcher.find()) {
                         matcher.appendReplacement(


GHX5T-SOL (approved):
Verified current head `7fabfcf6054a2b49236c802526d0f40a3a4c2a76`.

The change is scoped to the production `Replaced` buffer used by `Matcher.appendReplacement` / `appendTail`. Since `pom.xml` compiles with Java 17, the `StringBuilder` overloads are available, and the buffer is local to one mapped string operation. I also checked that this removes the only production `StringBuffer` occurrence; the remaining `StringBuffer` references are test coverage for APIs that accept it.

Local validation on this head:

- `mvn -q -DskipTests compile` -> passed.
- `mvn -q -Dtest=ReplacedTest -Djacoco.skip=true -Dhone.skip=true test` -> 11 tests, 0 failures, 0 errors, 0 skipped.
- `git grep -n "StringBuffer" origin/master -- src/main/java src/test/java` vs `HEAD` -> production `Replaced.java` occurrence removed; only test uses remain.
- `git diff --check origin/master...HEAD` -> clean.
- `git diff --no-ext-diff origin/master...HEAD | gitleaks stdin --no-banner --redact --exit-code 1` -> no leaks found.

Hosted Maven is currently unstable because the macOS job fails inside `hone-maven-plugin` with `Cannot run program "docker"`, and the other Maven jobs are cancelled. I don't see that as caused by this one-line `Replaced` change.

yegor256:
The change itself is simple and well justified — the buffer is local to a single lambda invocation, so there's no reason to pay for `StringBuffer`'s synchronization. I'm holding off on merging though because CI isn't fully green on the head commit: `mvn (macos-15, 23)` fails (looks like the hone-maven-plugin step can't reach Docker on that runner) and the ubuntu/windows `mvn` jobs were cancelled as a result. Doesn't look related to your diff, but I'd like a human to confirm before this merges.

Reassigning to yegor256 for manual review.

---

PR #1913, opened 2026-07-29
Fix #1912: rethrow original exception in ScalarWithFallback

### LINKED ISSUE #1912: ScalarWithFallback wraps the original exception instead of rethrowing it
In `src/main/java/org/cactoos/scalar/ScalarWithFallback.java:110`, when no registered `Fallback` supports the throwable, the private `fallback()` method throws `new Exception("No fallback found - throw the original exception", exp)`. It wraps the original instead of rethrowing it, so the message says the opposite of what the code does.

This contradicts the method's own Javadoc, which promises to "throw the original error if no fallback found" and documents `@throws Exception The original exception if no fallback found`. It matters because `value()` catches `Throwable`, so this branch is reached by everything the origin throws that no fallback matches. Given a single `Fallback.From<>(IOException.class, ...)`, an `IllegalArgumentException` from the origin leaves `value()` as a checked `java.lang.Exception` with the original merely as its cause: an unchecked exception silently becomes a checked one and the original type is no longer visible to the caller. A `StackOverflowError` is flattened the same way.

The fix is to rethrow `exp` in the `else` branch at line 110 rather than wrapping it: rethrow `RuntimeException` and `Error` as they are, and wrap only genuinely checked exceptions.

yegor256:
Closes #1912.

## Problem

In `ScalarWithFallback.fallback()`, when no registered `Fallback` supports the throwable, the `else` branch threw `new Exception("No fallback found - throw the original exception", exp)`. This *wrapped* the original throwable instead of rethrowing it — the message said the opposite of what the code did.

Because `value()` catches `Throwable`, this branch is reached by everything the origin throws that no fallback matches. The consequences:

- An unchecked exception (e.g. `IllegalArgumentException`) was silently turned into a checked `java.lang.Exception`, with the original only visible as the cause.
- The original exception type was no longer visible to the caller.
- An `Error` (e.g. `StackOverflowError`) was flattened the same way.

This also contradicted the method's own Javadoc, which promises to "throw the original error if no fallback found".

## Fix

`fallback()` now rethrows the original throwable instead of wrapping it:

- `RuntimeException` and `Error` are rethrown unchanged (type and identity preserved).
- Checked `Exception`s are rethrown directly.
- Only a raw `Throwable` that is neither `Exception` nor `Error` is wrapped (kept for type-safety, as the enclosing method is declared `throws Exception`).

## Tests

Added two tests to `ScalarWithFallbackTest` covering the issue's scenarios:

- `rethrowsOriginalUncheckedException` — an `IllegalArgumentException` from the origin, with only an `IOException` fallback, is rethrown as `IllegalArgumentException` (message preserved).
- `rethrowsOriginalCheckedException` — an `IOException` from the origin, with only an `IllegalArgumentException` fallback, is rethrown as `IOException` (message preserved).

All 11 tests in `ScalarWithFallbackTest` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014yf9dgnsJJqPtakMKmdCVP

---
_Generated by [Claude Code](https://claude.ai/code/session_014yf9dgnsJJqPtakMKmdCVP)_

### DIFF
--- a/src/main/java/org/cactoos/scalar/ScalarWithFallback.java
+++ b/src/main/java/org/cactoos/scalar/ScalarWithFallback.java
@@ -104,13 +104,18 @@ public final class ScalarWithFallback<T> implements Scalar<T> {
                     ).entrySet().iterator()
                 )
             );
-        if (candidates.hasNext()) {
-            return candidates.next().getKey().apply(exp);
-        } else {
-            throw new Exception(
-                "No fallback found - throw the original exception",
-                exp
-            );
+        if (!candidates.hasNext()) {
+            if (exp instanceof RuntimeException) {
+                throw (RuntimeException) exp;
+            }
+            if (exp instanceof Error) {
+                throw (Error) exp;
+            }
+            if (exp instanceof Exception) {
+                throw (Exception) exp;
+            }
+            throw new Exception(exp);
         }
+        return candidates.next().getKey().apply(exp);
     }
 }
--- a/src/test/java/org/cactoos/scalar/ScalarWithFallbackTest.java
+++ b/src/test/java/org/cactoos/scalar/ScalarWithFallbackTest.java
@@ -182,4 +182,38 @@ final class ScalarWithFallbackTest {
             new Throws<>(Exception.class)
         );
     }
+
+    @Test
+    void rethrowsOriginalUncheckedException() {
+        MatcherAssert.assertThat(
+            "Must rethrow the original unchecked exception when no fallback matches",
+            () -> new ScalarWithFallback<>(
+                () -> {
+                    throw new IllegalArgumentException("Unchecked failure");
+                },
+                new Fallback.From<>(
+                    IOException.class,
+                    exp -> "Fallback from IOException"
+                )
+            ).value(),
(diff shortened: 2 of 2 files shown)


yegor256:
CI is red on `mvn (macos-15, 23)`, but the failure is pre-existing on the base branch and unrelated to this change:

- The macos-15 job fails in the `hone-maven-plugin` `pull` goal with `java.io.IOException: Cannot run program "docker": error=2, No such file or directory` — the macOS runner has no Docker. This happens before any source is compiled, so this diff cannot trigger it.
- The `ubuntu-24.04` and `windows-2022` jobs show `cancelled` only because the matrix is fail-fast; they were stopped when macos-15 failed, not by anything in this PR.
- The same pattern (macos-15 failing on Docker, the other two cancelled) reproduces on the current `master` HEAD (`d8317ef`), and `master`'s `mvn` workflow has failed on every run since 2026-05-08 — long before this branch existed.

The fix itself (`ScalarWithFallback` rethrowing the original throwable) compiles and all 11 tests in `ScalarWithFallbackTest` pass locally. I'll re-check and re-run once the base branch's CI is green again.

---

